PR #1, opened 2018-06-13
Add API docs

calebeby:


docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+  }
+  name: string
+  district?: string
+  week?: number
+  // UTC date
+  startDate: string
+  // UTC date
+  endDate: string
+  // Only if user is logged in
+  starred?: false
```
fharding1: Didn't we decide we want to have a custom endpoint for starred since they're not really attached to the event but rather the user?

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+
+```ts
+type Data = {
+  redAlliance: string[]
+  blueAlliance: string[]
+  // UTC date - match predicted time
+  time: string
+  // example: qm3
+  key: string
+  starred?: false
```
fharding1: same as above

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+```
+
+# `/event/{eventKey}/match/{matchKey}/reports/{team}`
+
+## `PUT`
+
+### Request
+
+```ts
+type Data = {
```
fharding1: why so many different formats?

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+        match: string
+      }
+    | {
+        value: string
+        statName: string
+        // 2018orwil
+        event: string
+        // qm1
+        match: string
+      })[]
```
fharding1: shouldn't we also have general?

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
```
fharding1: general

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+}
+```
+
+
+### Response
+
+```ts
+type Data = {}
```
fharding1: I don't like this... the response body will be empty/null, not an empty object.

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+
+```ts
+type Data = {
+  redAlliance: string[]
+  blueAlliance: string[]
+  // UTC date - match predicted time
+  time: string
+  // example: qm3
+  key: string
+  starred?: false
```
fharding1: same

docs/docs.md
```diff
@@ -0,0 +1,628 @@
...
+        // total
+        attempts: number
+        // total
+        successes: number
+        statName: string
+      }
+}
+```
+
+# `/users/{userId}`
```
fharding1: needs comments about what admins can do vs normal users and what data admins get vs normal users

docs/docs.md
calebeby: Would it be empty or would it be `null`? empty is not valid JSON, but `null` is

docs/docs.md
calebeby: No u

docs/docs.md
fharding1: empty

docs/docs.md
```diff
@@ -0,0 +1,581 @@
...
+  username: string
+  name: string
+  admin?: true
+}
+```
+
+
+## `PUT`
+
+Only admins can modify other users
```
fharding1: except users can modify themselves

docs/docs.md
```diff
@@ -0,0 +1,581 @@
...
+### Response
+
+```ts
+type Data = null
+```
+
+
+## `DELETE`
+
+Only admins can delete other users
```
fharding1: users can delete themselves

docs/docs.md
```diff
@@ -0,0 +1,581 @@
...
+
+```ts
+type Data = null
+```
+
+# `/users`
+
+## `POST`
+
+Anyone can create a user. For admins, the user will be created
```
fharding1: for admins the users will be verified automatically, for non-admins or non-authenticated users the user will not be verified and will require admin approval

docs/docs.md
```diff
@@ -0,0 +1,581 @@
...
+immediately, for non-admins, the user won't be created immediately
+and it will be added to the user request queue
+
+### Request
+
+```ts
+type Data = {
```
fharding1: add isVerified

docs/docs.md
calebeby: Right, that was my intent when writing this

docs/docs.md
calebeby: That's why it says only admins can delete _other users_

docs/docs.md
fharding1: Seems vague to me.

docs/docs.md
fharding1: Seems vague to me.

calebeby:
Which endpoint should be used for verifying users, and what method

BrendanBurkhart:
I think either /authenticate or /login using POST

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+# `/users/{userId}/stars`
+
+## `GET`
+
+Anyone can view anyone's stars
+
+### Response
+
+```ts
+type Data = string[]
```
BrendanBurkhart: We should probably have a comment here saying that this is an array of eventkeys they have starred, otherwise it's a bit vague what 'stars' is.

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+  redAlliance: string[]
+  blueAlliance: string[]
+  // UTC date - match predicted time
+  time: string
+  // example: qm3
+  key: string
+}[]
+```
+
+# `/event/{eventKey}/star`
```
BrendanBurkhart: I think it would be a good idea to add a way to tell what endpoints need authentication or not - this method does since it needs to know which user sent the request but it would be nice to make that clearer.

docs/docs.md
fharding1: These comments are not very explicit. I think a better doc would be something like: "Admins can modify any property of any user expect for their ID. Users can modify themselves but not change whether they are an admin or whether they are verified. P.S. this needs verified.

docs/docs.md
fharding1: This formats kinda gross when you look at the markdown, they display on the same line so it's like "Anyone can modify themselves Only admins can modify other users"

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+      }
+    | {
+        attempted: boolean
+        succeeded: boolean
+        statName: string
+      })[]
+}[]
+```
+
+# `/event/{eventKey}/match/{matchKey}/stats`
```
BrendanBurkhart: I feel like this needs a comment explaining that it is total stats for those teams not just stats for that match - I got tripped up when I first saw this. It's pretty vague what the scope here is - is it stats for the match? For the team, but only within the event? Or stats from all the data we have on that team at all events?

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+```
+
+
+### Response
+
+```ts
+type Data = null
+```
+
+# `/event/{eventKey}/stats`
```
BrendanBurkhart: Is this stats on the teams performance at this event only, or at all events they have been to?

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+```
+
+# `/event/{eventKey}/team/{team}/stats/auto`
+
+## `GET`
+
+### Response
+
+```ts
+type Data = {
```
BrendanBurkhart: Does this get stats for all events the team has been to? I don't really see the point of that data since robots can change so much between events, and that is really not the data the endpoints says it's pointing to. But if its just stats for that event, the event data sent back is redundant. Also this seems like it would be better and more efficient to grab all reports for a team that every single stat split up over every single match. I know we need this data for graphs but it seems like throwing it all into one array would make it unnecessarily difficult to process.

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+        // qm1
+        match: string
+        attempted: boolean
+        succeeded: boolean
+        statName: string
+      })[]
+}[]
+```
+
+# `/event/{eventKey}/team/{team}/stats/teleop`
```
BrendanBurkhart: It seems redundant to send which event the data is coming from since the endpoint already specifies that. As with the auto version of this, having all the data in one giant array seems like a not-great way to represent it.

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+  webcasts: {
+    type: "twitch" | "youtube"
+    url: string
+  }[]
+  location: {
+    name: string
+    lat: number
+    lon: number
+  }
+  name: string
```
BrendanBurkhart: Are we going to do the shortname/fullname thing again or just go with one of them?

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+```
+
+# `/team/{team}/stats/teleop`
+
+## `GET`
+
+### Response
+
+```ts
+type Data = {
```
BrendanBurkhart: Why does /teleop send data for both teleop and auto? Comparing this endpoint (`/team/{team}/stats/teleop`) to `/event/{eventKey}/team/{team}/stats/teleop`, the /event one specifies what event each data point comes from but this one doesn't which seems backwards to me. Also, what is the use case for grabbing all the data on a team across all events?

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+        attempts: number
+        successes: number
+        statName: string
+      }
+    | {
+        attempted: boolean
+        succeeded: boolean
+        statName: string
+      })[]
+  auto: (
```
BrendanBurkhart: I think this data needs to include auto mode name

docs/docs.md
```diff
@@ -0,0 +1,585 @@
...
+```
+
+# `/event/{eventKey}/match/{matchKey}/info`
+
+## `GET`
+
+### Response
+
+```ts
+type Data = {
```
BrendanBurkhart: There is no way to get score from a match here, is that intentional?

docs/docs.md
calebeby: I think we'll just do shortname everywhere

docs/docs.md
calebeby: At this point the frontend UI doesn't display match scores, but that will probably be added at some point so I added it to the API

docs/docs.md
calebeby: :heavy_check_mark:

docs/docs.md
calebeby: :heavy_check_mark:

docs/docs.md
calebeby: :heavy_check_mark:

docs/docs.md
calebeby: This event only :heavy_check_mark:

docs/docs.md
calebeby: event_ team_ auto

Yeah they can get detailed stats for either just that single event or for all events

calebeby: event_ team_ auto

docs/docs.md
calebeby: :heavy_check_mark:  on the removing event part, since it is duplicated

I think a giant array is fine

PR #6, opened 2018-06-21
Docs Feedback

calebeby:


src/api.ts
```diff
@@ -279,8 +295,12 @@ export const getStarredEvents = (userId: number) =>
 // Anyone can modify themselves
 // Only admins can modify other users
 export const modifyUser = (userId: number, user: EditableUser) =>
-  putRequest<null>(`/users/${userId}`, user)
+  patchRequest<null>(`/users/${userId}`, user)
```
BrendanBurkhart: Is this patching with the full user? Because that basically defeats the purpose of the patch, and makes it a put.

calebeby: No, it doesn't have to be a full user. I forgot to switch this, cause there was a bug in the docs generator. I will add that.

docs/docs.md
```diff
@@ -491,14 +513,17 @@ Anyone can view any user
 ```ts
 type Data = {
   username: string
-  name: string
+  name: {
```
fharding1: I don't really like this... can we just do firstName and lastName?

docs/docs.md
```diff
@@ -508,9 +533,15 @@ Only admins can modify other users
 ```ts
 type Data = {
```
fharding1: Everything here needs to be optional...

src/api.ts
```diff
@@ -1,6 +1,6 @@
 type PeregrineResponse<T> =
```
fharding1: {"data": {}, "error": {"error": "bla", "code": 4}}

docs/docs.md
calebeby: :heavy_check_mark:

src/api.ts
calebeby: :heavy_check_mark:

docs/docs.md
calebeby: :heavy_check_mark:

PR #64, opened 2018-10-01
Set up preact devtools in development

calebeby:


calebeby:
Oops I will need to fix that

PR #67, opened 2018-10-02
Update user-related docs

fharding1:


src/api.ts
```diff
@@ -319,4 +312,4 @@ export const deleteUser = (userId: number) =>
 
 // Response is the JWT
```
calebeby: You can delete this comment since now there is a `jwt` property

PR #75, opened 2018-10-08
Add create event and create match to api

### LINKED ISSUE #74: Add POST routes for events and matches to the docs

fharding1:
Closes #74

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -209,7 +209,7 @@ type Data = {
 type Data = {
   redAlliance: string[]
   blueAlliance: string[]
-  // UTC date - match predicted time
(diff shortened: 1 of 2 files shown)


docs/docs.md
```diff
@@ -201,6 +201,31 @@ type Data = {
...
 
+## `PUT`
+
+### Request
+
+```ts
+type Data = {
+  redAlliance: string[]
+  blueAlliance: string[]
+  // UTC date - match predicted time
```
calebeby: Can you make this comment say scheduled time? You can override the property on the object using a union, even if the value is the same and it will update the comment

PR #93, opened 2018-10-15
Fix docs event district

calebeby:
I realized that I needed `EventInfo` to have a full district name, not only the abbreviation.


calebeby:
Yes I moved it in the other PR because I thought it would help Rollup with code splitting, but it ended up not affecting it and I forgot to put it back

PR #130, opened 2018-11-03
Add realms

brendanburkhart:


calebeby:
Can we have some sort of indication on an event whether it is global or specific to that realm?

docs/docs.md
```diff
@@ -495,6 +510,100 @@ type Data = {
...
+
+## `GET`
+
+Public realms will be returned. If logged-in, the user's realm will also be returned.
+
+### Response
+
+```ts
+type Data = {
+  ID: number
```
calebeby: Is this supposed to be capitalized?

calebeby:
```suggestion
  id: number
```

PR #217, opened 2018-12-31
Add refresh tokens

calebeby:
You can test this by verifying that `refreshToken` is created in localstorage, and that if you delete the `jwt`, it is automatically re-populated by making a request to `/refresh`

https://refresh-tokens--peregrine.netlify.com/


src/jwt.ts
```diff
@@ -10,19 +11,42 @@ interface JWT {
   raw: string
 }
 
+const isExpired = (jwt: JWT) => new Date(jwt.exp * 1000) < new Date()
```
fharding1: I think this should be <= and you should compare by casting `new Date()` to a number.

```
  The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.  The
   processing of the "exp" claim requires that the current date/time
   MUST be before the expiration date/time listed in the "exp" claim.```

calebeby: You can compare dates without casting to numbers

fharding1: I know, but if the expiry time is specified as a unix time it only specifies the precision to a second, and a date has more precision than that.

calebeby: Won't it just expire up to a second early on the FE?

fharding1: Yeah

calebeby: Is that a problem?

fharding1: I mean it's an off-by-one. It's not a huge problem, but it's technically wrong and it's easy to fix, so IMO you might as well fix it.

PR #227, opened 2019-01-07
Default info tab, add upcoming match to info tab

fharding1:
Fixes:
* Analysis table centering on large monitors

Features:
* Default to info tab on the event page
* Upcoming match on info tab


src/components/page/index.tsx
```diff
@@ -15,14 +15,16 @@ type Props = {
   children?: ComponentChildren
   back?: string | (() => void)
   tabs?: Tab[]
+  defaultTab?: string
```
calebeby: Why is this a prop, instead of just rearranging the tabs? 🤔

calebeby:
capture

It is unclear that this is the next match. It looks like a bug because there are 2

src/routes/event/index.tsx
```diff
@@ -66,129 +66,146 @@ const Event = ({ eventKey }: Props) => (
...
-                                icon={webcastIcon(w)}
-                                href={w}
-                                target="_blank"
-                                rel="noopener"
-                              />
-                            ))
+    renderSuccess={({ matches, eventInfo, eventStats, schema }) => {
+      const now = new Date()
+      const nextMatch = matches
+        ? matches.find(m => m.time !== undefined && m.time > now) || matches[0]
```
calebeby: It should not display a "next match" if all matches are in the past
I don't think we need a `|| matches[0]`

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
calebeby: Can you delete this plz

fharding1: https://i.redd.it/hs5bpm5y5s121.jpg

src/routes/event/index.tsx
```diff
@@ -66,129 +66,146 @@ const Event = ({ eventKey }: Props) => (
...
-                                key={w}
-                                icon={webcastIcon(w)}
-                                href={w}
-                                target="_blank"
-                                rel="noopener"
-                              />
-                            ))
+    renderSuccess={({ matches, eventInfo, eventStats, schema }) => {
+      const now = new Date()
+      const nextMatch = matches
```
calebeby: Instead of

```js
matches ? matches.... : undefined
```

Can you do

```js
matches && matches....
```

src/routes/event/index.tsx
```diff
@@ -66,129 +66,145 @@ const Event = ({ eventKey }: Props) => (
...
+                  ) : (
+                    <Spinner />
+                  )}
+                </div>
+              ),
+            },
+            {
+              name: 'Matches',
+              contents:
+                matches === undefined ? (
```
calebeby: All arrays including `[]` are truthy, can you rearrange the cases here back to be similar to how it was, and check for `!matches` instead of `matches === undefined` to display the spinner?

PR #279, opened 2019-03-07
Simple event sorting

fharding1:


src/routes/home/index.tsx
```diff
@@ -4,22 +4,26 @@ import Page from '@/components/page'
...
-            <Spinner />
-          )}
-        </div>
-      )}
+      renderSuccess={({ events }) => {
+        events = events && events.sort(compareEvents)
+        return (
+          <div>
+            {events ? (
+              events.map(e => (
```
calebeby: Can you inline the `.sort` here? saves bytes :fire: :money_mouth_face:

src/utils/compare-events/index.ts
```diff
@@ -0,0 +1,29 @@
+interface PartialEvent {
+  startDate: Date
+  endDate: Date
+}
+
+/**
+ * Compares 2 events. Returns negative if a comes first, positive if b comes first
```
calebeby:
```suggestion
 * Compares 2 events by date. Returns negative if a comes first, positive if b comes first
```

src/utils/compare-events/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param a First Event
+ * @param b Second Event
+ */
+export const compareEvents = (a: PartialEvent, b: PartialEvent) => {
```
calebeby: Can we make it so it sorts it like this:

If a.week and b.week are the same and not `undefined`, sort them alphabetically

Otherwise, use the logic you already have

I could add that in a follow up PR if you want

src/utils/compare-events/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  const now = new Date()
+
+  const aDiff = Number(now) - Number(a.endDate)
```
calebeby: Can you put the `const now` outside of this method, and change it to `new Date().getTime()`

Then it doesn't have to fetch the date and coerce it to number each time in the loop

src/utils/compare-events/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  if (aDiff < 0 && bDiff >= 0) {
+    return -1
+  }
+
+  if (bDiff < 0 && aDiff >= 0) {
+    return 1
+  }
+
+  return Math.abs(Number(now) - Number(a.startDate)) <
```
calebeby: you don't have to coerce now to a number here since it already is

fharding1: Oof yeah

PR #370, opened 2019-05-25
Improve tests

calebeby:
I forked preact-testing-library because they weren't keeping it up to date with react-testing library and it was missing a bunch of stuff. Then I refactored some tests to stop using the hacky stuff that the old preact-testing-library required (`await nextTick()`). I also added some more tests for components.

I also improved form validation for the register and login routes (submit button is disabled if the form is invalid, text boxes are marked as required but they are only red if you have focused them)


src/cache/index.ts
```diff
@@ -36,7 +36,8 @@ export const transaction = <T = void>(
         const handleResult = handler(tx.objectStore(storeName))
         if (handleResult) {
           ;(handleResult as IDBRequest<T>).onsuccess = event => {
-            resolve(((event.target as unknown) as { result: T }).result)
+            const { result } = (event.target as unknown) as { result: T }
+            if (result !== undefined) resolve(result)
```
calebeby: Unrelated to the rest of the PR but this fixes a bug where it would look for nonexistent items in the cache, and resolve with undefined even though the value doesn't exist in the cache

PR #485, opened 2019-08-09
GitHub actions

calebeby:
its not working :frowning:


calebeby:
it was deleted. github doesn't allow me to use the yml syntax, so I'd have to create the workflows using the old deprecated syntax. I'll wait till the public release, so I can use the modern yml syntax

PR #526, opened 2019-09-08
Add match creation ability

### LINKED ISSUE #209: Add ability to add/edit matches

calebeby:
http://create-matches--peregrine.netlify.com/

If you go to an event page, and you are logged in as an admin there will be an event admin button. Click it and then you can create matches. Please don't add matches to events other than peak performance. It sets the match date/time to the current time, because time/date editors are hard :man_shrugging:

Closes #209


calebeby:
The back end PR has a breaking change that will break this

fharding1:
Ah ok. I just deployed that.

PR #533, opened 2019-09-12
Add logout button

brendanburkhart:
Add logout functionality, and add logout and login button to menu.


src/components/menu.tsx
```diff
@@ -48,11 +52,17 @@ const menuItemStyle = css`
...
-const MenuItem = ({ href, children, icon }: MenuItemProps) => {
-  const isActive = resolveUrl(href) === window.location.href
+
+const MenuItem = ({ href, children, icon, onClick }: MenuItemProps) => {
+  const isActive = href ? resolveUrl(href) === window.location.href : false
+
   return (
     <li>
-      <a class={clsx(isActive && activeStyle, menuItemStyle)} href={href}>
+      <a
```
calebeby: It shouldn't be an `<a>` if it doesn't have an `href`. It should be an `<a>` when href is supplied, `<button>` otherwise. I'll go ahead and make this change

calebeby: Oh I think you mentioned that in slack yesterday lol. I'll fix the css too

calebeby:
Thanks Brendan!

PR #535, opened 2019-09-14
Add offline reports saving/syncing

calebeby:
When you load a page on peregrine
2 seconds after the page loads it is supposed to attempt to upload any/all offline reports.

Every 30 seconds, it is supposed to attempt to upload any/all offline reports.

When it attempts to upload any/all offline reports
Any that still fail stick around in localstorage.

When a report is submitted,
If it fails,
It saves the report to localstorage.

When the offline reports queue is not empty
A menu item should appear:

screenshot

That page has a list of offline reports. You can't interact with them, but you can see that they exist and you can force it to attempt an upload (rather than waiting 30 seconds).

screenshot

When you test this please only submit reports to https://offline-reports--peregrine.netlify.com/events/2019iscmp so we don't mess up the data for teams we may compete with.

Testing URL: https://offline-reports--peregrine.netlify.com/events/2019iscmp


calebeby:
@brendanburkhart @fharding1 When you merge this, *Squash and merge* to `dev`

PR #581, opened 2019-10-11
iOS fixes

calebeby:
Works around webkit bugs cuz apple is bad

- Analysis table should scroll correctly in iOS now
- Fixes the boolean chart scroll indicators for iOS
- Fixes iOS bug on info group card

https://ios-fixes--peregrine.netlify.com/events/2019cc

### DIFF
--- a/src/components/chart.tsx
+++ b/src/components/chart.tsx
@@ -351,8 +351,8 @@ const Chart: FunctionComponent<ChartProps> = memo(
         )}
 
         {lerpedPoints.map((y, x) => (
+          // eslint-disable-next-line caleb/react/jsx-key
           <circle
-            key={x} // eslint-disable-line caleb/react/no-array-index-key
             cx={xLerper(x)}
             cy={y}
             class={pointStyle}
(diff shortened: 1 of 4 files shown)


src/components/chart.tsx
```diff
@@ -506,17 +510,21 @@ const BooleanChart: FunctionComponent<ChartProps> = ({
...
-        ))}
+      <div
+        class={innerBooleanChartWrapperStyle}
+        ref={elementRef}
+        onScroll={recomputeScrolling}
+      >
+        <div class={innerBooleanChartStyle}>
+          {points.map((p, i) => (
+            <BooleanDisplay
+              key={i} // eslint-disable-line caleb/react/no-array-index-key
```
fharding1: Isn't it not great to use the index key since React won't be able to uniquely identify elements?  I guess it doesn't matter too much since once they're rendered we never change/re-order them (like sort or w/e). Plus it's not introduced via this PR. Maybe a comment clarifying why this is fine would be good?

fharding1:
P.S. so glad this is fixed :fire: :100:

PR #584, opened 2019-10-12
Add stat rank position to analysis table

### LINKED ISSUE #573: Add stat rankings to analysis table
This is a feature request we've gotten from people on the team as well as Chad Stryker from Steel Ridge:

The Steel Ridge scouting team really enjoyed using Peregrine at the Oregon Girls Gen event.  The lead scouter told me that this was, by far, the easiest scouting method we have ever used.  Also, since we only had three scouters at the event, being able to share data with your team made it possible for us to develop an effective alliance during the play off rounds.
I would also like to provide some feedback to the Peregrine development team.  While examining the analysis table to evaluate specific teams, I found myself having to repeatedly sort by a column and then manually count rows to determine the rank of the team for the given performance attribute.  It would be very useful if the results table automatically displayed a rank column to make that task easier.  See the attached example.

Peregrine Rank Column Example

fharding1:
Closes #573

### DIFF
--- a/src/components/analysis-table.tsx
+++ b/src/components/analysis-table.tsx
@@ -153,8 +153,8 @@ const teamNumCellStyle = css`
 const teamRankStyle = css`
   position: absolute;
   font-size: 0.75rem;
-  left: 5%;
-  top: 5%;
+  left: 0.15rem;
+  top: 0.15rem;
   color: ${grey};
 `
 
--- a/src/components/analysis-table.tsx
+++ b/src/components/analysis-table.tsx
@@ -153,7 +153,7 @@ const teamNumCellStyle = css`
 const teamRankStyle = css`
   position: absolute;
   font-size: 0.75rem;
-  left: 0.15rem;
+  right: 0.15rem;
   top: 0.15rem;
   color: ${grey};
 `
--- a/src/components/analysis-table.tsx
+++ b/src/components/analysis-table.tsx
@@ -150,6 +150,13 @@ const teamNumCellStyle = css`
   }
 `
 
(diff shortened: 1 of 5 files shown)


src/components/table.tsx
```diff
@@ -71,7 +71,7 @@ const sortButtonStyle = css`
 export interface Column<CellType, RowType> {
   /** Column name, used in title row */
   title: string
-  renderCell: (cellValue: CellType) => JSX.Element
+  renderCell: (cellValue: CellType, cellIndex: number) => JSX.Element
```
calebeby: Should the 2nd parameter be called `rowIndex`?

fharding1: ah, yes

src/components/analysis-table.tsx
```diff
@@ -150,6 +150,14 @@ const teamNumCellStyle = css`
   }
 `
 
+const teamRankStyle = css`
+  position: absolute;
+  font-size: 0.75rem;
+  left: 5%;
```
calebeby: Can you use `rem`s plz?

fharding1: ye

src/components/table.tsx
```diff
@@ -199,14 +199,16 @@ const TableRow = memo(
   <RowType extends any>({
```
calebeby: There is no longer a benefit for `TableRow` to be `memo(`'d anymore because its props change on pretty much every sort. Let's remove `memo(`.

fharding1: Good point

PR #590, opened 2019-10-17
Use gmapsUrl from TBA rather than lat/lon directly

brendanburkhart:
We are currently plugging event lat/lon directly into Google Maps, this doesn't show the name or details of the event space. TBA provides full Google Maps URLs for events, using this instead is a bit nicer.

### DIFF
--- a/src/api/event-info/index.ts
+++ b/src/api/event-info/index.ts
@@ -7,6 +7,8 @@ export interface EventInfo {
   // district "display_name" from TBA
   fullDistrict?: string
   locationName: string
+  lat?: number
+  lon?: number
   gmapsUrl?: string
   key: string
   // the ID of the realm the event belongs to
--- a/src/api/event-info/index.ts
+++ b/src/api/event-info/index.ts
@@ -7,8 +7,7 @@ export interface EventInfo {
   // district "display_name" from TBA
   fullDistrict?: string
   locationName: string
-  lat?: number
-  lon?: number
+  gmapsUrl?: string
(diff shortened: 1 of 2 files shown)


src/api/event-info/index.ts
```diff
@@ -7,8 +7,7 @@ export interface EventInfo {
   // district "display_name" from TBA
   fullDistrict?: string
   locationName: string
-  lat?: number
```
calebeby: You can keep these on here, right? Cuz they still exist on the response, right?

brendanburkhart: I suppose, I can't really think of anything they're used for though, and once this is fully deployed we could remove them from the BE - I left them in the BE response so we could have a smooth transition. Do you want to keep them around for something in the future?

calebeby: It is being used for sorting events by distance when 2 events are at the same time, if they both have locations.

brendanburkhart: alright, adding back now. How did it pass the CI linter/typecheck?

calebeby: I had the same question so I dug into it. It passes types because the event comparer essentially requires "2 objects that have start/end dates and optionally lat/lon" (because many offseasons don't have locations). An `EventInfo`, without `lat/lon`, satisfies that.

It's a bummer but I guess it is what is expected from the type system. It _works_ without `lat`/`lon` so the types don't jump out to tell you that it's not going to work.

PR #597, opened 2019-10-22
Add event search/filtering

### LINKED ISSUE #574: Add event search/filtering
From the home page, users should be able to type in a search box to filter the event list. Events whose `key`, `name`, `district`, `fullDistrict`, or `locationName` matches the entered text should be shown.

If you don't want to do the visual styles for the search box, that's ok, I can do it for you.

marek-jarab:
Closes #574


calebeby:
Thanks @marek-jarab, this works really well! I'm looking through your code now, and I'll leave some feedback

calebeby:
@marek-jarab Can you run `npm run type` and fix the type errors that that shows?

marek-jarab:
I have fixed these type errors which occurred after "npm run type" command.

src/routes/home.tsx
```diff
@@ -5,27 +5,55 @@ import Spinner from '@/components/spinner'
...
 
 const now = new Date()
-
 const Home = () => {
   const events = useEvents()
   const location = useGeoLocation()
+  const [term = '', setTerm] = useState(0);
+
+  function searchingFor(term: string | number) {
+    return function (x: any) {
```
calebeby: I'd suggest not using the `any` type, because we can use a more accurate type. Also renaming the variable would be helpful.

```js
    return function (event: ProcessedEventInfo) {
```

You will also need to add:

```js
import { ProcessedEventInfo } from '@/api/event-info'
```

src/routes/home.tsx
```diff
@@ -5,27 +5,55 @@ import Spinner from '@/components/spinner'
...
 import { useGeoLocation } from '@/utils/use-geo-location'
 import { useEvents } from '@/cache/events/use'
+import { useState } from 'preact/hooks'
 
 const now = new Date()
-
 const Home = () => {
   const events = useEvents()
   const location = useGeoLocation()
+  const [term = '', setTerm] = useState(0);
```
calebeby: Why is the initial value 0? Can it be '' instead?
```suggestion
  const [term, setTerm] = useState('');
```

src/routes/home.tsx
```diff
@@ -5,27 +5,55 @@ import Spinner from '@/components/spinner'
...
+
+  function searchingFor(term: string | number) {
```
calebeby: I think it would make sense for this to be always passed a string.

```suggestion
  function searchingFor(term: string) {
```

src/routes/home.tsx
```diff
@@ -5,27 +5,55 @@ import Spinner from '@/components/spinner'
...
+        return (x.name.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+          (x.key.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+          (x.locationName.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+          (x.district.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+          (x.fullDistrict.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term);
+      }
+
+      return (x.name.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+        (x.key.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+        (x.locationName.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term);
```
calebeby: I think these nested conditionals could use some simplification. Here are some suggestions:

- You don't need to call `.toString()` on things that are already strings. If you are using VSCode, you can hover over a value (like `x.key`), and it will show you the type. If it is already `string`, calling `.toString()` is unnecessary.
- You are repeating `term.toString().toLowerCase()` many times. I'd suggest extracting this to a variable. Also we don't need to use `.toString` once we remove `number` from the type:
  ```js
  const lowercaseQuery = term.toLowerCase();
  ```
- You have `|| !term` on every single condition. I think it would be much simpler if we used a single early return at the start of the function instead. This would still handle the case of if the user has not typed anything, in which case we want to show all matches:
  ```js
  if (!term) return true;
  ```

src/routes/home.tsx
```diff
@@ -5,27 +5,55 @@ import Spinner from '@/components/spinner'
...
   const location = useGeoLocation()
+  const [term = '', setTerm] = useState(0);
+
+  function searchingFor(term: string | number) {
+    return function (x: any) {
+      // because not all objects contain properties 'district' and 'fullDistrict' you have to check if particular object contains property 'distict' and if not tell method to not use it. Otherwise this error will occur (Uncaught (in promise) TypeError: Cannot read property 'toString' of undefined)
+      if (x.hasOwnProperty('district')) {
+        return (x.name.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+          (x.key.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+          (x.locationName.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
```
calebeby: `event.key` and `event.locationName` exist regardless of whether the `district` property exists, so I don't think they should be within this `if` statement that checks whether `district` exists.

src/routes/home.tsx
```diff
@@ -5,27 +5,55 @@ import Spinner from '@/components/spinner'
...
+          (x.fullDistrict.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term);
+      }
+
+      return (x.name.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+        (x.key.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term) ||
+        (x.locationName.toString().toLowerCase().includes(term.toString().toLowerCase()) || !term);
+
+    }
+  }
+  function searchHandler(event: any) {
```
calebeby: I think `setTerm` should be called directly in the inline arrow function in the `onInput` attribute.

src/routes/home.tsx
marek-jarab: But I think that, then you will not be able to filter events correctly. For example : 
You have some event which has "property: district" and you are trying to find this event according his "property:key" .
So you type key into search bar ,but this event will not be displayed beacuse you are not checking if event.key match with what you typed into search bar. You checking it only in case if object has not property "district".
Or am I wrong? I also tried to remove it from if statement in code and it doesnt work then.

calebeby: You are right. I think there is a different way of handling this that allows us to not duplicate these conditions. I'll suggest it after you apply the other suggestions.

src/routes/home.tsx
marek-jarab: When I tried to  defined type event: ProccessedEventInfo. There is a problem with properties district and fulldistrict. This error occurred. App  running correctly anyway, but i cannot figure out how to solve this issue . Do you have any suggestion? 
event

calebeby: Yeah. I think if you do:

```js
if (
  event.district !== undefined &&
  event.district.toLowerCase().includes(lowercaseQuery)
) {
  return true;
}

if (
  event.fullDistrict !== undefined &&
  event.fullDistrict.toLowerCase().includes(lowercaseQuery)
) {
  return true;
}
```

Then that should hopefully fix the type errors and also that allows us to not have to duplicate checking name/key/locationName (because it only returns if it matches)

marek-jarab: Yes it works. Thanks for advice

marek-jarab:
I think I have succesfully implemented all your suggestions. I committed them so you can check it now.

src/routes/home.tsx
```diff
@@ -5,27 +5,47 @@ import Spinner from '@/components/spinner'
...
 import { useEvents } from '@/cache/events/use'
+import { useState } from 'preact/hooks'
+import { ProcessedEventInfo } from '@/api/event-info'
 
 const now = new Date()
-
 const Home = () => {
   const events = useEvents()
   const location = useGeoLocation()
+  const [term = '', setTerm] = useState('');
```
calebeby:
```suggestion
  const [term, setTerm] = useState('');
```

You can remove the default value because you are passing an initial value to `useState`

src/routes/home.tsx
```diff
@@ -5,27 +5,47 @@ import Spinner from '@/components/spinner'
...
   return (
     <Page name="Home" back={false}>
+      <div>Search bar</div>
+      <div class="searchBar">
+        <input type="text" onInput={(e: any) => setTerm(e.target.value)} />
+      </div>
       <div>
         {events ? (
           events
+            .filter(searchingFor(term))
```
calebeby: @marek-jarab Can you inline the `searchingFor` function here?

calebeby:
@marek-jarab Can you run `npm run lint` which will automatically run prettier to format the files? That will fix the `CI / lint` check

src/routes/home.tsx
marek-jarab: What do you mean by that? because when I run "npm run lint" it organized rows as they used to be.
rows

calebeby: I mean that instead of defining the `searchingFor` function outside and calling it here, can you turn it into an inline arrow function here to filter the events?

marek-jarab: So you want to  remove current function searchingFor() and create new arrow function inside return method with the whole logic? Won't the return() method be  pointlessly too complicated? Also it's seem for me quite complicated to implement , as I am not used to work with arrow functions often

calebeby: Here is what it would look like:

```js
            .filter(event => {
              if (!term) return true
              return (
                event.name.toLowerCase().includes(lowerCaseQuery) ||
                event.key.toLowerCase().includes(lowerCaseQuery) ||
                event.locationName.toLowerCase().includes(lowerCaseQuery) ||
                (event.district !== undefined &&
                  event.district.toLowerCase().includes(lowerCaseQuery)) ||
                (event.fullDistrict !== undefined &&
                  event.fullDistrict.toLowerCase().includes(lowerCaseQuery))
              )
            })
```
I don't think it makes the `return` too complicated

marek-jarab: Oh Ok. I thought it would be better to create function outside the return method, but this is also good approach.

calebeby:
@marek-jarab I'm going to merge this PR to a new `event-filtering` branch, where I'll add CSS to the search box, then I'll merge it to `dev`

Thanks for taking the time to implement this, good job :tada: :tada:

marek-jarab:
Thanks. I also want to thank you for your suggestions and advices. Again I had opportunity to learned something new :)

calebeby:
Yeah, no problem! Let me know if you ever want to contribute again!

PR #621, opened 2019-11-03
Simplify router and fix unmount bug

calebeby:


src/components/spinner/index.tsx
```diff
@@ -87,7 +87,8 @@ const Spinner = () => {
   // prevents spinner from showing when loading is < 100ms
   // makes user experience less jarring
   useEffect(() => {
-    setTimeout(() => setIsShown(true), 100)
+    const t = setTimeout(() => setIsShown(true), 100)
+    return () => clearTimeout(t)
```
calebeby: Here was the bug. It is a minor bug. The timeout was sometimes firing after the component unmounted, because I didn't have a cleanup to clear the timeout.

src/router.tsx
```diff
@@ -14,20 +14,13 @@ interface Route {
...
-  [url: string]: AnyComponent
-}
-
 const routers: ((url: string) => void)[] = []
 
 export const route = (url: string) => {
   routers.forEach(router => {
     router(url)
   })
+  history.pushState(null, '', url)
```
calebeby: I refactored this component because I was pretty sure the bug was in here. It turned out the bug was not in here, but my refactorings made it smaller, more correct, and simpler, so I kept them

PR #732, opened 2020-01-17
Get events from all years from the BE

brendanburkhart:


src/utils/move-focus-inside/index.ts
```diff
@@ -4,7 +4,6 @@ const tababblesStr = tabbables.join(',')
 
 export const moveFocusInside = (el: HTMLElement) => {
   if (el.matches(tababblesStr)) return el.focus()
-  // TODO: Should this handle elements with modified tabindex, like https://github.com/theKashey/focus-lock/blob/master/src/utils/tabOrder.js
```
calebeby: Can you undo this change?

PR #733, opened 2020-01-17
Upgrade to node.js 13.6.0

fharding1:
Netlify will need to be updated, too.


calebeby:
@fharding1 `.nvmrc` is the source of truth. That needs to be updated. Also why should we be on node 13?

calebeby:
(For context)

We decided to wait till node 14 because it has a much longer lifecycle

PR #798, opened 2020-02-23
Match videos

calebeby:
https://match-videos--peregrine.netlify.com/events/2020week0/matches/qm14

https://match-videos--peregrine.netlify.com/events/2019cc/matches/qm10

- Resize your browser really small
- Go to devtools network tab and emulate offline. The video player should go away instantly
- Go to devtools network tab and emulate a 3g or slower network. If you reload the page, the video player should not load until you click on the black box (conditional lazy loading)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1377,6 +1377,12 @@
         "@types/yargs": "^13.0.0"
       }
     },
+    "@mdi/js": {
+      "version": "4.9.95",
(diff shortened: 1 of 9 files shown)


calebeby:
@brendan can you test on ios

brendanburkhart:
It doesn’t load the video player until I click on it (I’m on WiFi, speed should be fine), but other than that seems to work as expected.

calebeby:
That is expected because Safari doesn't support the API to get the connection info (other then online/offline) ✅

PR #818, opened 2020-03-01
Add rank to analysis table

calebeby:
screenshot

https://analysis-table-rank--peregrine.netlify.com/events/2020wasno/analysis

Test out sorting by rank.

Also verify that sorting by rank works correctly on match pages

Also I changed it so that the little number next to the team number on the analysis table is only shown if you are sorting by a real stat. Because if you are sorting by team number, it is unnecessary to know "team ____ is the nth-lowest team number", and if you are sorting by rank, it is just duplicating the information in the rank column.


brendanburkhart:
I can’t sort lowest to highest on columns other than team number and rank - clicking the header once sorts by that column, clicking more times does nothing

calebeby:
oof good catch

calebeby:
I'm seeing that on dev as well though. I'll create a separate PR to fix that

brendanburkhart:
sgtm, thanks

PR #820, opened 2020-03-02
Fix wonky match cards

calebeby:
I was seeing this weird layout for the match cards specifically on the match page on my phone:

image

Now it is fixed

https://fix-wonky-match-cards--peregrine.netlify.com/events/2020wasno/matches/qm1


calebeby:
@brendanburkhart I can only reproduce it on my phone, not on my computer with small screen, I don't know why

PR #821, opened 2020-03-02
Fix reverse sort

calebeby:
https://fix-reverse-sort--peregrine.netlify.com/events/2020orore/analysis

The bug was introduced in #808


calebeby:
ignore CI, it is bad

it is intermittently forgetting how to install node

PR #905, opened 2020-04-18
Add geolocation button

### LINKED ISSUE #890: Improve geolocation prompt experience
Right now it appears after a fixed amount of time, which often makes it seem like it is popping up at a random time. We can come up with a better solution for this (maybe opt-in for geolocation prompt) to improve this

calebeby:
Closes #890

https://add-geolocation-button--peregrine.netlify.app/

- There should be no geolocation prompt until you click the geolocation button on the homepage
- The geolocation button should not appear if you already clicked "allow" or "block"
- When you click on the geolocation button the browser should prompt for your location

In iOS and safari, we don't have the ability to know if the user has clicked "allow" or "block" so the button will always appear.

### DIFF
--- a/src/routes/home.tsx
+++ b/src/routes/home.tsx
@@ -58,7 +58,13 @@ const Home = () => {
       <div class={filterStyle}>
         <Dropdown options={years} onChange={setYear} value={year} />
         <TextInput onInput={setQuery} label="Search for Events" />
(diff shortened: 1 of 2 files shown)


calebeby:
@j10k11 Can you test in iOS again? Now it won't prompt until you tap the button

PR #1027, opened 2020-07-05
Non-Existent Events

j10k11:
Added messages that would appear for event pages that are non-existent.


src/routes/event.tsx
```diff
@@ -52,44 +54,65 @@ const noMatchesStyle = css`
 const Event = ({ eventKey }: Props) => {
   const matches = useEventMatches(eventKey)
   const eventInfo = useEventInfo(eventKey)
-  const newestIncompleteMatch = matches && nextIncompleteMatch(matches)
-
+  const newestIncompleteMatch = isData(matches) && nextIncompleteMatch(matches)
+  console.log(matches)
```
calebeby, line 58: We can remove this `console.log`

```suggestion
```

src/routes/event.tsx
```diff
@@ -52,44 +54,65 @@ const noMatchesStyle = css`
...
   return (
     <Page
       name={eventInfo?.name || <code>{eventKey}</code>}
```
calebeby, line 61: This line is where the page title is displayed. Right now, if there is a network error or a non-network error, the page title gets displayed as `Error` or `NetworkError`. Can you change this to use isData so that if there is an error it displays the event key?

src/routes/event.tsx
```diff
@@ -52,44 +54,65 @@ const noMatchesStyle = css`
...
-        </Heading>
-        {eventInfo && <EventInfoCard event={eventInfo} />}
-        <Button href={`/events/${eventKey}/analysis`}>Analysis</Button>
-      </div>
+      {eventInfo instanceof NetworkError
+        ? 'Could not connect to server. Please check your connection.'
+        : eventInfo instanceof Error && (
+            <Fragment>
+              This event does not exist.
+              <Button href={`/`}>Existing Events</Button>
```
calebeby, line 70: Since there aren't any expressions embedded in the template literal, we can change this to regular quotes and remove the `{}`

```suggestion
              <Button href="/">Existing Events</Button>
```

src/routes/event.tsx
```diff
@@ -52,44 +54,65 @@ const noMatchesStyle = css`
...
+                <EventMatches matches={matches} eventKey={eventKey} />
+              ) : (
+                <p class={noMatchesStyle}>No matches yet</p>
+              )
+            ) : matches ? (
+              matches instanceof NetworkError ? (
+                'Network error. Please check your connection.'
+              ) : (
+                matches instanceof Error && (
+                  <Fragment>This Event Does Not Exist.</Fragment>
```
calebeby, line 107: I think we can use a string here instead of a fragment

calebeby:
@j10k11 Can we center these views? It looks like they are being added to one of the grid columns.

image

image


Also can we consolidate this into a single column:

image

src/routes/event-match.tsx
```diff
@@ -98,14 +98,18 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
         m.group +
         (m.num ? ' Match ' + m.num : '') +
         ' - ' +
-        (event ? event.name : eventKey)
+        (isData(event) ? event.name : eventKey)
```
calebeby, line 101: @j10k11 I think we can leave out the changes to this file so Victor can handle them in his PR

calebeby (dismissed):
@j10k11 This looks great! I left my first round of PR feedback

src/routes/event.tsx
```diff
@@ -49,47 +58,105 @@ const noMatchesStyle = css`
...
-      name={eventInfo?.name || <code>{eventKey}</code>}
+      name={isData(eventInfo) ? eventInfo.name : <code>{eventKey}</code>}
       back="/"
-      class={eventStyle}
+      class={
+        eventInfo instanceof Error
+          ? eventInfo instanceof NetworkError
+            ? networkErrorStyle
+            : eventErrorStyle
+          : eventStyle
```
calebeby, line 98: We can simplify this so that the ternaries aren't nested:

```suggestion
        eventInfo instanceof NetworkError
          ? networkErrorStyle
          : eventInfo instanceof Error
          ? eventErrorStyle
          : eventStyle
```

src/routes/event-match.tsx
calebeby: Just a reminder about this feedback. Victor is taking care of all the changes to the match page in his PR, so you can undo the changes here.

calebeby:
This is looking really good! Great job with the layout work for when the errors are displayed!

I have a little bit of feedback about the padding/margin/spacing used for the errors:

- Can we remove the `errorSectionStyle` and put the error elements one level higher? You can re-add the space between the errorSectionStyle and the "existing events" button using `grid-gap`
- Can you combine the `eventErrorStyle` and `networkErrorStyle` css blocks into just one? They seem to be 90% the same (I could be missing something, there could be a reason why they have to be separate)
- I can't tell if the media queries in `eventErrorStyle` and `networkErrorStyle` are being used for anything. Can they be removed? (remember they are used for applying css properties just for certain screen sizes)
- Can we make the "existing events" button be less wide? I think just the width of the text is fine

I think this is almost ready for merge!!!

varCepheid:
There are a lot of errors being thrown in here and I'm not sure I can fix all of them. I think, like the non-existent matches PR, I should leave this alone for now since it isn't a critical issue.

varCepheid:
I fixed all the errors, but I think I may have undone the changes that Jennifer made. I'll keep working on this.

varCepheid:
The only change I see left is the content of the message. I see something previously about an "Existing Events" button. Do you think what I have is okay?

ev118:
Looks good, I changed a bit of the text but nothing else had major issues.

PR #1054, opened 2020-07-25
Don't display analysis table if there is no data

### LINKED ISSUE #1029: Don't display analysis table if there is no data
If you go to the event analysis page for an event with no data: https://peregrine.ga/events/2020micmp/analysis

We currently display an empty analysis table. Instead we should display something like "No data" or "No data for this event". This can be displayed in a small card or just on the background

Here is where the analysis table is used: https://github.com/Pigmice2733/peregrine-frontend/blob/dev/src/routes/event-analysis.tsx#L40

We can check if `eventStats` is empty and if it is, then display the fallback text

FaytheLL:
closes #1029 
Make sure this event shows no data:
https://deploy-preview-1054--peregrine.netlify.app/events/2020award/analysis
Make sure this event shows data:
https://deploy-preview-1054--peregrine.netlify.app/events/2020waspo/analysis


calebeby:
@FaytheLL You can merge this pull request by clicking "Squash and merge". Then after a few minutes the changes will be live on the dev site: https://dev.peregrine.ga/

PR #1057, opened 2020-07-30
Added a message for non-existent matches

VictorHuynh:
Hey there! Not sure if this is what I'm supposed to be doing, but hopefully it's right. My code's a bit sketch but it gets the job done. 
So far, I've:
- Added a message for non-existent matches
   - Has a button that returns to the event page
   - Can determine if match key is parseable and displays the match # or "this match" accordingly

There are a couple of errors, yet the code runs anyways. Please critique!

[EDIT for bed9084 🛏️ ]
- Changed it so that invalid match keys return `null` when put through the `formatMatchKey` function
- Displays "This Match Does Not Exist" when parsed match key returns `null`
- Made some formatting changes (shout out to the Prettier extension)
   - General removal of unnecessary code and alterations for clarity

[EDIT for 1e0e78d]
It's been a while, but I finally did it! 😭🎉 For this commit, the following changes were made:
- Optimized the code
- Undid previous (and questionable) changes


src/routes/event-match.tsx
```diff
@@ -91,6 +94,36 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
     [match],
   )
 
+//  Variables used to assist in deciding if match key is parse-able 
+const parsedMatchType = m.group.replace(/\d+/g, '')
+const matchKeyFinalChar = m.group.charAt(m.group.length - 1)
+const numbers = ["1","2","3","4","5","6","7","8","9","0"]
```
calebeby, line 100: I think we had talked about this before, but can you remind me why it is beneficial to determine whether a match key is parseable?

How do we handle match keys that are parseable but are nonexistent matches? (i.e. `qm1000`)

VictorHuynh: I'm more in favor of Solution 3. Coming from a user's perspective, they may find it difficult to interpret the error message if only a match key is provided for context. With that said, I'd **immensely** appreciate any suggestions that'll simplify the code (it has always been a bit convoluted to begin with 💫)

calebeby: Sounds good :+1: 

I think it makes the most sense to change the `formatMatchKey` function so that it no longer allows for invalid match keys. You can make it so that it returns `null` if the match key is invalid. I would change [this line](https://github.com/Pigmice2733/peregrine-frontend/blob/match-page-errors/src/utils/format-match-key/index.ts#L6) to return just `null` rather than the key in uppercase. Then in this file you can do:

```js
const parsedMatch = formatMatchKey(m.key)
```

and

```js
parsedMatch === null ? "This match does not exist" : /* ... */
```

Not completely sure this suggestion will work as-is, but it should at least be a good starting point. I have some free time later this week in case we want to do a call to work through this.

src/routes/event-match.tsx
```diff
@@ -91,28 +96,62 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
         (event ? event.name : eventKey)
       }
       class={matchStyle}
     >
       <Button href={`/events/${eventKey}/matches/${matchKey}/scout`}>
         Scout Match
       </Button>
-      {match ? <MatchCard match={match} eventKey={eventKey} /> : <Spinner />}
-      {match && matchHasBeenPlayed && (
+      {isData(match) ? (
```
calebeby, line 143: Here it is impossible for `isData(match)` to be false, because we already checked on line 99, and if it was false, we would have returned already.

src/components/analysis-table.tsx
```diff
@@ -30,7 +30,7 @@ import { EventTeamInfo } from '@/api/event-team-info'
 interface Props {
   eventKey: string
   teams: ProcessedTeamStats[] | undefined
-  schema: Schema
+  schema: Schema | Error
```
calebeby, line 33: I think that it would be better to avoid passing in an error as the schema to the analysis table. We can instead wrap the place where we are using the analysis table, with a check to make sure the schema exists before rendering.

If you make that change, then you should be able to undo all of the changes in this file

src/utils/use-promise/index.tsx
```diff
@@ -4,11 +4,11 @@ import { CancellablePromise } from '@/utils/cancellable-promise'
 import { NetworkError } from '@/api/base'
 
 export const usePromise = <T extends any>(
-  promiseCreator: () => Promise<T> | T,
+  promiseCreator: () => Promise<T> | T | false,
   dependencies: any[] = [promiseCreator],
 ) => {
   const emitError = useErrorEmitter()
-  const [val, setVal] = useState<T | undefined | Error | NetworkError>(
+  const [val, setVal] = useState<T | undefined | Error | NetworkError | false>(
```
calebeby, line 11: I'm not sure I understand what the `| false` is for here. I don't remember, I might have suggested you to make this change :joy: but I don't think this is the best option here. I think that `promiseCreator` should return `undefined` if it can't get the data.

I think we should undo the changes to this file, which will probably cause errors in other files, but I think that would be a better change.

src/routes/event-match.tsx
```diff
@@ -61,19 +63,22 @@ const showEventResults = 'Event Results'
 
 type SelectedDisplay = typeof showMatchResults | typeof showEventResults
 
+// eslint-disable-next-line complexity
```
calebeby, line 66: I don't think this eslint rule is triggering on the next line, so I think we can take this out

src/routes/event-match.tsx
```diff
@@ -91,28 +94,58 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
 
+  if (!isData(match)) {
+    return (
+      <Page
+        back={`/events/${eventKey}`}
+        name={
+          (m === null
+            ? matchKey + ' - '
+            : m.group + (m.num ? ' Match ' + m.num : '') + ' - ') +
+          (event ? event.name : eventKey)
```
calebeby, line 105: Can we make this a variable so we don't have to duplicate it between this section and line 131?

src/routes/event-match.tsx
```diff
@@ -91,28 +94,58 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
+        }
+        class={matchStyle}
+      >
+        <Card class={matchStyle}>
```
calebeby, line 109: It is probably safer to create a new css block specifically for this, rather than reusing `matchStyle`. This branch doesn't have the redesigned match page styles yet, but if it did, I think it might break the layout of this card.

src/routes/event-match.tsx
```diff
@@ -91,28 +94,58 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
+          <Icon icon={alert} />
+          <div
+            class={css`
+              font-size: 2rem;
+              text-align: center;
+            `}
+          >
+            {m === null
+              ? 'This Match Does Not Exist'
+              : m.group + (m.num ? ' Match ' + m.num : '') + ' Does Not Exist'}
```
calebeby, line 119: Can we differentiate between a network error and a server error and a 404?

If I block the request to the peregrine API in devtools and clear the cache, then it shows "Qual 1 does not exist", but can we change it to say "Unable to load Qual 1" in the case that there was a network error in fetching it?

varCepheid: From what I see, if the whole thing isn't throwing a huge error, it has to be a 404. It has to get the matchKey from somewhere to run at all. If it has the matchKey and still return null, it's a 404 error; if it doesn't have the matchKey, it's probably a network or server error, but I don't think I could run anything in here to detect that.

varCepheid: Never mind, I was looking in the wrong place. All the way up at L97, it's checking whether or not match is defined. I spent a couple hours trying to understand what would allow match to be defined or undefined, and I again don't understand the software quite well enough to figure that out.

varCepheid: I attempted to address this in my new commit. As I mentioned, I couldn't test it because of weird errors I didn't create.

varCepheid:
@calebeby The tests just aren't working. I don't understand them quite well enough to figure out why they're failing. I see one error in utils/cancellable-promise/index about the type of the `resolve` argument of the `execute` function, but that isn't showing up on the tests. Any ideas?

calebeby:
@varCepheid I don't remember enough about this PR off the top of my head to be able to tell you easily. We can meet to go over this PR and get it back up and running after competition.

This PR fixes a bug that only happens if you type random stuff into the URL bar, which is a relatively uncommon case so I don't think we need to worry about this before competition.

varCepheid:
@calebeby I cannot verify what state these changes are in because I can't see any events with matches. When I go to the page for an event that has matches, it throws this error.
image
The only line of code I changed is in line 93 in `components/match-card` where I changed `Fragment` to `preact.Fragment` (this also required importing preact, line 8) because it was throwing an error around it. I don't know exactly what `Fragment` and `El` are, so I don't understand what this error means or why it's occuring.

varCepheid:
Caleb, the reason so many files were changed is because they had errors that won't pass the typecheck if I didn't fix them. Because the branch this is merging to isn't updated to `dev`, the issues haven't been fixed upstream. If we merge this, I can fix them in a merge commit from `dev` to `handle-network-errors`.

calebeby:
@varCepheid I went ahead and merged dev into handle-network-errors and handle-network-errors into this branch (we don't need PR's when merging in that direction). The changes were pretty small so I'm not sure that there were changes that affected lint passing :thinking:

varCepheid:
There are errors that arise when I use the original code on these files. I don't know why the same errors don't exist elsewhere, but the typecheck won't go through if they're not here.

calebeby:
I think your editor may be showing lint errors that aren't actual problems, that are already disabled in the project config.

calebeby:
If you run `npm run eslint` and it passes and `npm run type` and it passes, then it is fine. Your editor may not be configured correctly to use the eslint setup.

`npm run lint` passes on the `match-page-errors` branch so most of these changes are not needed.

calebeby:
If you run `npm run eslint -- --report-unused-disable-directives` then it will list the `eslint-disable` comments that are not needed.

calebeby:
When you say "the typecheck won't go through", what specific command are you running (or seeing in your editor)?

calebeby:
From our slack conversation:

> Here is test code I used to detect which kind of error was being sent:
> ```js
> if (m instanceof Error) {
>   if (m instanceof ServerError) {
>     return <h1>Server error: {m.message}</h1>
>   }
>   if (m instanceof NetworkError) {
>     return <h1>Network error: {m.message}</h1>
>   }
>   return <h1>Other error: {m.message}</h1>
> }
> ```
> You can do something similar for the match page to display a more helpful error depending on which kind of error happened

PR #1207, opened 2020-10-31
Cache users offline

### LINKED ISSUE #1097: When you log out, clear indexeddb cache and user-specific localstorage values
[Here](https://github.com/Pigmice2733/peregrine-frontend/blob/dev/src/jwt.ts#L53) is the code that handles what happens when you log out. Currently, if you are logged in and have access to information that is only visible _because_ you are logged in, if you log out, the information could still be visible because it could be cached. This can be fixed by clearing the caches when a user logs out

### LINKED ISSUE #1098: Cache user ids/firstname/lastname
It would be useful to be able to see users' names even while offline. Currently there is no caching of user names, so if you for example save a report offline and then view it, it shows as "anonymous" since the user name cannot be fetched. If that was cached, we could display the actual name instead of "anonymous"

calebeby:
Closes #1098 
Closes #1097


brendanburkhart (approved):
🚢

Issue #1099, opened 2020-08-23
When you log in directly from the home page, after you log in, it goes back to where you were before you went to peregrine


calebeby:

calebeby:
It might make sense to have a `?back=https://` query parameter like most sites use

PR #1237, opened 2020-11-22
Fixed login bug that redirected the user to the wrong page when logging in (Issue 1099)

PepperLola:
Fixed the login bug that redirected you to the page you were on before peregrine ([issue 1099](https://github.com/Pigmice2733/peregrine-frontend/issues/1099)).


src/routes/login.tsx
```diff
@@ -1,6 +1,13 @@
 import Authenticated from '@/components/authenticated'
 
-const goBack = () => window.history.back()
+const useQuery = () => {
+  return new URLSearchParams(window.location.search);
+}
+
+const goBack = () => {
+  // eslint-disable-next-line caleb/react-hooks/rules-of-hooks
+  window.location.href = decodeURIComponent(useQuery().get("from") || '');
```
calebeby, line 9: I'm curious why useQuery is a separate function, does it provide any advantage here? It looks like ESLint thinks that it is a react hook (because its name starts with `use`) but as far as I can tell it's not.

src/routes/login.tsx
```diff
@@ -1,6 +1,13 @@
 import Authenticated from '@/components/authenticated'
 
-const goBack = () => window.history.back()
+const useQuery = () => {
+  return new URLSearchParams(window.location.search);
```
calebeby, line 4: In this project we've been using [`qss`](https://github.com/lukeed/qss#qssdecodequery) to encode/decode query strings. It might make sense to switch to the built-in `URLSearchParams` eventually, but for now can we continue to use `qss` for consistency?

calebeby (changes requested):
This is great! I left a couple comments but it is almost ready to go!

Let me know if you have any questions, great job figuring this out :tada:

calebeby:
@PepperLola the `lint` check is failing because some files are not matching `prettier` auto-formatting. You can run `npm run lint` to fix them. Also you can enable the prettier extension in your editor so it will do that automatically when you save files.

src/components/menu.tsx
```diff
@@ -218,7 +218,12 @@ export const Menu = ({ onHide, visible }: Props) => {
               </MenuItem>
             ) : (
               <>
-                <MenuItem icon={login} href="/login">
+                <MenuItem
+                  icon={login}
+                  href={`/login?from=${encodeURIComponent(
+                    window.location.href,
```
calebeby, line 224:
```suggestion
                    location.pathname,
```

It should still work, but with this change the query parameter will look a little bit nicer without the domain

calebeby (approved):
Looks great! Left 1 more comment, feel free to merge once you address it

Make sure to merge using the "Squash and merge" option

calebeby (approved):

PR #1240, opened 2020-11-25
Switched icons to use material design icons directly (Issue #1234)

OpioidSalt:
Deleted icons folder and switched icons to use mdi icons.


calebeby (changes requested):
@OpioidSalt :tada: thanks for doing this!

A couple bits of feedback:
1. Can you combine multiple imports to `@mdi/js` in the same file into a single import? If you do `npm run lint` it will do this automatically, but then it will leave behind extra empty lines that you'll need to remove
2. Were there not any icons that were custom / not from material design icons? I think I remember making one in inkscape but I could be wrong

calebeby (approved):

PR #1266, opened 2020-12-18
Fix issue with analysis table width on match page

Jas0nTran:
Analysis tables now don't stretch horizontally on wide screens.


src/routes/event-match.tsx
```diff
@@ -172,6 +172,9 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
               class={css`
                 overflow-y: hidden;
                 grid-area: analysisTable;
+                @media screen and (width > 118.75rem) {
```
calebeby, line 175: Can you do `max-width: 100%` to make it use the scrollbar instead of using a media query? The analysis table does not have a fixed width and it changes year-to-year and also a little bit based on font rendering, etc.

src/routes/event-match.tsx
```diff
@@ -171,10 +171,10 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
             <Card
               class={css`
                 overflow-y: hidden;
```
calebeby, line 173: I don't think we need the overflow-y afaict

varCepheid:
The changes Jason made look like they would work, but I don't have a wide screen to test on. The computer I'm using now can't make itself any wider. Caleb, if you still have access to a wide screen, can you check if it works?

calebeby:
@varCepheid Yes it is working ✅ 

Here is a trick you can use to simulate a different screen size:

- In Chrome, open devtools (right click on page, then click inspect)
- Click the "toggle device toolbar" button near the top left
- In the dimensions dropdown choose "Responsive"
- Edit the dimensions to whatever

Screen Shot 2022-01-21 at 8 44 03 PM

calebeby:
@varCepheid I think we can remove the overflow-y as I mentioned in my previous review and then we can merge this

calebeby (approved):

PR #1481, opened 2022-01-20
set comment max width

varCepheid:
10rem as a test, subject to change


varCepheid:
I seem to be having a lot of issues with npm. I don't know how all of these tests work or if they're important, but I definitely didn't cause all the errors it's listing.

Issue #1480, opened 2022-01-17
Report page stretches to full width if comment is long


calebeby:
https://dev.peregrinefrc.com/reports/3484

![screenshot](https://user-images.githubusercontent.com/13206945/149809000-ba8fa19a-f819-4f13-b66c-da58efb741fc.png)

It looks funny. We should set a `max-width` on the card so that it does not stretch so much.

Here is the file: https://github.com/Pigmice2733/peregrine-frontend/blob/dev/src/routes/report.tsx

PR #1482, opened 2022-01-21
Max width for Report Viewer

varCepheid:


calebeby:
@varCepheid I'm not sure what this fixes, the report viewer appears to me the same as before: 
screenshot
https://deploy-preview-1482--peregrine.netlify.app/reports/3484.

Could you please include a description/testing instructions in your PR so that I know what it fixes?

calebeby:
Also to be clear, if this is intended to fix #1480, we want the entire report to be narrowed down, not just the comment

src/components/report-viewer.tsx
```diff
@@ -111,14 +113,22 @@ export const ReportViewer = ({ report, onEditClick }: Props) => {
...
       {report.comment && (
-        <CommentCard
-          report={report}
-          showReporter={false}
-          linkToReport={false}
-        />
+        <div
+          class={css`
+            max-width: 15rem;
+          `}
```
calebeby, line 119: @varCepheid Can we apply the max-width directly to the `<Card>` component that contains the `<ReportViewer>` [here](https://github.com/Pigmice2733/peregrine-frontend/blob/dev/src/routes/report.tsx#L66)

That way if the field names happen to be longer then the comment will still be the full width of the report viewer card, which would fix this:

Screen Shot 2022-01-21 at 8 56 47 PM

Also, I think a little wider than 15rem might be good so it does not feel quite so squished, maybe 20rem or 25rem?

calebeby (changes requested):

varCepheid:
That fix it?

calebeby:
@varCepheid Yep, works great, thanks :tada: 

I think [this change](https://github.com/Pigmice2733/peregrine-frontend/pull/1482/files#diff-dbe4442faaeafea3ad9c8ea58707c63ca9408361308f664c3dbd25c71c9b1ab0R26) is the only change needed to fix the bug, the rest of the changes aren't needed. Do you remind removing the max-width from the comment (since we have a max-width set on the report viewer card, which is sufficient), and also removing the wrapper `<div>`s you added?

src/routes/report.tsx
```diff
@@ -16,11 +16,14 @@ const reportPageStyle = css`
   padding: 2rem;
   justify-content: center;
 `
+
+// defines qualities of the report area (white background) on the page
```
calebeby, line 20: In UI design we usually refer to something like this as a card. Tends to have white background, shadow, rounded corners, etc. That's what I was trying to describe with the name of the variable. Do you have a suggestion for a better variable name that we could use here?

calebeby:
@varCepheid I appreciate the addition of the comments, some of them are really helpful to explain/summarize what is happening in the code near them.

Generally, I try to use comments when the code is not very clear what it is doing, when the code unexpectedly deviates from a normal way of doing things, and when there are not opportunities to simplify the code to make it explain itself.

Here are some of the comments that you added that are really helpful:

- https://github.com/Pigmice2733/peregrine-frontend/blob/max-width-for-comments/src/components/report-viewer.tsx#L81
   Since the links "interrupt" the "____ in ____ at ____" sentence, and the link syntax is kinda verbose, this comment summarizes what is going on, which makes it more skimmable, and there isn't really an opportunity to simplify the code itself
- https://github.com/Pigmice2733/peregrine-frontend/blob/max-width-for-comments/src/components/report-viewer.tsx#L96
  Since the component is called `TeamPicker`, that could confuse a reader of the code because there is not actually a way to choose a team here. The name of that component originated in the report editor, where a scout would use that component to select which team they were scouting. The comment clarifies that and it is helpful. If you wanted to, I think there is an opportunity to rename the component to something that better encompasses its two uses (as a team picker for scouting, and as a list of teams in alliances in a match). But I think leaving the name as-is and adding a comment is fine too.

Some of the other comments are less helpful (to me at least):
- https://github.com/Pigmice2733/peregrine-frontend/blob/max-width-for-comments/src/routes/report.tsx#L52
   The name of the `<Page>` component is intended to convey that it is the root page component and contains the page layout. Does the comment communicate/clarify something?
- https://github.com/Pigmice2733/peregrine-frontend/blob/max-width-for-comments/src/routes/report.tsx#L14
   In this codebase, we use variable names ending with `Style` to indicate that they are CSS style blocks. So the variable name `reportPageStyle` is intended to communicate that it is a CSS block that will be applied to the report page. And if we scroll down, we can see that it is applied to the `<Page>` root component in the `<ReportPage>` component.

I won't nitpick each of the comments you added one-by-one, but those are the kinds of things I think about when deciding whether to write comments and what to communicate in comments.

Let me know if you have any questions, I don't mean for this to come across as harsh and I don't want to discourage you! Comments are important and valuable!

calebeby:
@varCepheid Forgot to mention:

I'm sure you've noticed the annoying syntax limitation of comments within JSX (HTML within JS/TS)
While adding the comments next to/between attributes like this works:

```js
<div class="something" /* some comment */>
```

There is a syntax that is more common and I prefer slightly (unless the comment is specifically referring to one of the attributes):

```js
{/* some comment */}
<div class="something">
```

calebeby:
Also there are a bunch of changes to `package-lock.json`, usually this happens when one person is using a different version of `npm`. Can you install the latest version of `npm` by running `npm i -g npm`, and then run `npm i` again to hopefully switch the `package-lock.json` back to `"lockfileVersion": 2`?

calebeby (approved):

Issue #1191, opened 2020-10-17
Don't stretch analysis table horizontally on wide screens


calebeby:
![screenshot](https://user-images.githubusercontent.com/13206945/96352436-f6a39380-1077-11eb-81d6-c10213501c6d.png)

https://dev.peregrinefrc.com/events/2020waspo/matches/qm1

PR #1483, opened 2022-01-21
Analysis table max width

varCepheid:


calebeby:
@varCepheid Is this supposed to be for https://github.com/Pigmice2733/peregrine-frontend/issues/1191? If so, that issue already has a PR linked to it, so we don't need this 2nd PR. If you want to take over that PR, can you coordinate with @Jas0nTran? I think that PR was 99% done, only my most recent comment on that PR needed to be addressed and should be a 1-line change.

https://github.com/Pigmice2733/peregrine-frontend/pull/1266

PR #1485, opened 2022-01-30
loading match pages offline

### LINKED ISSUE #1224: Match pages fail to load offline
If you go to a match page (e.g.  online, go offline, and return to/reload the match page, it will be stuck on the loading page. @calebeby thinks this is due to the analysis page. I can't currently fully repro this on dev - the match info and score panels still load, but on the latest deploy preview nothing loads whatsoever besides the loading indicator.

### LINKED ISSUE #1484: video covers analysis table when offline
image
https://dev.peregrinefrc.com/events/2019orwil/matches/qm2
Hiding the videos offline should fix this.

varCepheid:
closes #1224
closes #1484

### DIFF
--- a/src/routes/event-match.tsx
+++ b/src/routes/event-match.tsx
@@ -25,6 +25,10 @@ import { getReports } from '@/api/report/get-reports'
 import Icon from '@/components/icon'
 import { mdiPlus } from '@mdi/js'
 import { createShadow } from '@/utils/create-shadow'
+import {
+  ConnectionType,
+  useNetworkConnection,
+} from '@/utils/use-network-connection'
 
 interface Props {
   eventKey: string
@@ -97,15 +101,24 @@ type SelectedDisplay = typeof showMatchResults | typeof showEventResults
 
 // eslint-disable-next-line complexity
 const EventMatch = ({ eventKey, matchKey }: Props) => {
+  // checks for no network connection
+  const netConn = useNetworkConnection()
+  const isOnline = netConn !== ConnectionType.Offline
+
   const m = formatMatchKey(matchKey)
   const event = useEventInfo(eventKey)
   const match = useMatchInfo(eventKey, matchKey)
-  const reports = usePromise(
-    () => getReports({ event: eventKey, match: matchKey }),
-    [eventKey, matchKey],
-  )
-  const schema = useSchema(event?.schemaId)
-  const teams = usePromise(() => getEventStats(eventKey), [eventKey])
+  const reports = usePromise(() => {
+    if (isOnline) {
+      return getReports({ event: eventKey, match: matchKey })
+    }
+  }, [eventKey, matchKey, isOnline])
+  const schema = useSchema(isOnline ? event?.schemaId : undefined)
+  const eventTeamsStats = usePromise(() => {
+    if (isOnline) {
+      return getEventStats(eventKey)
+    }
+  }, [eventKey, isOnline])
 
   const [selectedDisplay, setSelectedDisplay] = useState<SelectedDisplay>(
     showEventResults,
@@ -119,16 +132,15 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
     if (matchHasBeenPlayed) setSelectedDisplay(showMatchResults)
   }, [matchHasBeenPlayed])
 
-  const teamsStats = usePromise(
-    () =>
-      match &&
-      Promise.all(
+  const teamsStats = usePromise(() => {
+    if (match && isOnline) {
+      return Promise.all(
         [...match.redAlliance, ...match.blueAlliance].map((t) =>
           getMatchTeamStats(eventKey, match.key, t).then(processTeamStats),
         ),
-      ),
-    [match],
-  )
+      )
+    }
+  }, [match, isOnline])
 
   return (
     // page setup
@@ -169,7 +181,7 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
               </Card>
             )}
           </div>
-          {schema && (
+          {schema && isOnline && (
             // card including the analysis table and tabs for match/event data
             <Card
               class={css`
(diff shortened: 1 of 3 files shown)


varCepheid:
Everything worked well for me for that one page. Last fix is to make the card that the text appears on a little wider and the text bold.

varCepheid:
Tested on that one match page as well as another, it's working how I think it should.

src/utils/use-network-connection.ts
```diff
@@ -64,7 +67,10 @@ export const useNetworkConnection = () => {
     const connection = navigator.connection
     const updateConnectionStatus = () => {
       if (navigator.onLine) {
-        if (connection?.effectiveType === '4g') {
+        if (
+          connection?.effectiveType === '4g' &&
+          (connection.type === 'wifi' || connection.type === 'ethernet')
```
calebeby, line 72: @varCepheid This doesn't seem to work quite right, when I am on my laptop on wifi, the video doesn't load automatically. `navigator.connection.effectiveType` is `4g` on my computer and `navigator.connection.type` is not defined. https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/type#browser_compatibility. It looks like that property is not supported on most computers (but it is on phones, except iphones). Is there different logic we can use when connection.type is not defined?

varCepheid: Simplest fix I see is running only the first condition when `connection.type` is undefined. Any issue with that? It fixes the issue on my computer, but I don't have a good way to test at the moment on a phone.

calebeby: 👍 Sure, works for me. But that means that on iphones it will always load by default 😬 I guess that is fine 🤷

src/routes/event-match.tsx
```diff
@@ -229,9 +246,28 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
+            >
+              <p>
+                <em>
+                  <b>
+                    Analysis table, videos, and reports are only available
+                    online.
+                  </b>
+                </em>
+              </p>
+            </Card>
```
calebeby, line 270: @varCepheid Great job with adding this card to inform the user why stuff is missing. I have some feedback on the visual design and text:

- I think it looks better without the italics
- Can we move it to the top above the match info card? Also I think it would look good to have the same width as the other cards.
- Can we add an icon in front of the message? We have an `<Icon>` component to display an icon and you could use [this icon](https://materialdesignicons.com/icon/cloud-off-outline) (or another one)
- What do you think about changing the text like this (the cloud emoji would be the slashed cloud icon):
   :cloud: **No connection** Showing limited information
- Can we use consistent padding all the way around the element? The `<p>` tag is adding extra margin above/below (we don't need a `<p>` tag here).

calebeby:
@varCepheid Great work with this, it is looking really good.

- I think changing the 2nd line of text to "Showing limited information" would be a little easier to read than "Analysis table, videos, and reports are only available online." For some reason, when I read "Analysis table, videos, and reports are only available online" my brain reads it as meaning "only Analysis table, videos, and reports are available online" and then my brain has to parse it another time to understand it right.
- The cloud icon and the "no connection" text are not aligned, it would be nice for the midline of the text to be aligned with the middle of the icon. You can use flexbox or grid to implement this (happy to help/provide more details if you want)
   screenshot
- Generally, unless there is semantic meaning associated with bolded text or semantic breaks in phrasing, `<b>` and `<br>` should be avoided. Those elements came from an era when the HTML elements themselves were concerned with visual display more than semantic meaning. With HTML 5, the meaning of those elements has been adjusted, and in cases like this where we just want bolded text and a split line (no semantic meaning), it is better to use elements without semantic meaning like `<div>` and `<span>` with styling applied in CSS. Do you mind switching to div and/or span?

src/routes/event-match.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable caleb/@typescript-eslint/restrict-plus-operands */
```
calebeby, line 1: I think this eslint-disable comment is no longer needed.

src/routes/event-match.tsx
```diff
@@ -39,6 +44,35 @@ const blueStyle = css``
...
+  grid-template-columns: auto;
+  grid-template-areas:
+    'topArea'
+    'leftColumn';
+  padding: 0.75rem;
+  @media (max-width: 930px) {
+    grid-template-columns: 1fr;
+    grid-template-areas:
+      'topArea'
+      'leftColumn';
```
calebeby, line 69: This grid-template-areas property is the same as the one outside of the media query, so there is no need to override it here.

src/routes/event-match.tsx
```diff
@@ -142,15 +189,33 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
         <>
+          {!isOnline && (
+            <div class={offlineDisplayInfo}>
+              <Card
+                class={css`
+                  justify-self: center;
+                  padding: 1rem;
+                  text-align: center;
+                `}
+              >
```
calebeby, line 210: Would it work to combine the outer div and the card into just the card?

```js
<Card class={offlineDisplayInfo}>
```
And move the CSS properties from the inline block up to the offlineDisplayInfo style block?

calebeby: On screen sizes < 540px and > ~500px, the edges of the offline card don't line up with the edges of everything else. I think combining the outer div with the card may fix that, but in case it doesn't, can you fix that as well?

calebeby:
Also, your approach is totally valid but I am curious whether it would simplify the CSS to move the offline stuff into the left column div. There may be a reason why that won't work that I haven't thought of, but just curious

varCepheid:
I don't think it would simplify. It makes sense to me to leave that piece alone and not possibly throw in a card that we don't always want to display.

src/routes/event-match.tsx
```diff
@@ -90,22 +102,54 @@ const matchWithVideoStyle = css`
...
+  }
+`
+
+const offlineDisplayInfo = css`
+  grid-area: topArea;
+  display: grid;
+  grid-template-areas:
+    'iconArea title'
+    'detail detail';
+  justify-self: center;
```
calebeby, line 121: Can you add `align-items: center` so that the "No connection" text is vertically centered with the icon?

calebeby:
The width difference here doesn't look great IMO. Mind switching it back to match the width? (adding it to the leftColumn might be the easiest way but I could be wrong)

screenshot

calebeby:
Can you use grid-gap on the no connection card to add an equal gap between all the items, and also add padding to the card, and remove the padding on the individual children of the card?

calebeby (approved):
Great work! Thanks for putting up with all my nitpicky feedback :wink: 

Feel free to merge

PR #1487, opened 2022-02-08
Main page all events

### LINKED ISSUE #1486: display some set of events on main page all times of year
Right now, when I open the website, it tries to display 2022 events. Since those obviously don't exist, it just doesn't display anything. I think it should either display the last year of events (2021) or all events.

varCepheid:
closes #1486

### DIFF
--- a/src/components/report-viewer.tsx
+++ b/src/components/report-viewer.tsx
@@ -59,6 +59,10 @@ const fieldValuesStyle = css`
   }
 `
 
+const reportCommentCard = css`
+  max-width: 10rem
+`
+
 export const ReportViewer = ({ report, onEditClick }: Props) => {
   const reporterId = report.reporterId
   const eventInfo = useEventInfo(report.eventKey)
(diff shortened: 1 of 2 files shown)


varCepheid:
This was the easiest way I saw to fix the issue. It probably wouldn't be that much harder to do all events, but since it narrows to 20, it doesn't make a huge difference.

calebeby:
@varCepheid I think we can solve this without fetching a whole second unused list of events by adjusting the default year logic here:
https://github.com/Pigmice2733/peregrine-frontend/blob/dev/src/routes/home.tsx#L50-L52

That code calls `useYears` to fetch a list of all the years which have events, and the `useQueryState` hook is being passed the current year as the default year. Maybe we could adjust the logic to use the most recent year in the `years` array, rather than the current year?

calebeby:
@varCepheid Brendan imported the 2022 events so this bug won't appear again until next year. But we can of course still go ahead with this PR (once you get to my feedback).

I did not think through the timing with this PR when I asked Brendan to import the 2022 events, now this bug is harder to reproduce :slightly_frowning_face: . You can re-reproduce this issue if you change the line `const currentYear = now.getFullYear()` in the `home.tsx` route to `const currentYear = 2023` so it thinks that the current year is a year that doesn't have events

calebeby (approved):
screenshot

Love it when a bugfix ends up with more lines deleted than added.

Great work!

varCepheid:
Can I merge this, then?

PR #1489, opened 2022-02-10
displayed a message if the event has not started

varCepheid:


varCepheid:
@calebeby There's still a red-flag popup that says "no schema found". I tried to fix it the way we did for the offline-match-pages issue, but that didn't solve it. Any ideas?

calebeby:
@varCepheid That is because we have not defined a schema for 2022 events. We can do that after FRC publishes the data structure that they will include for 2022 matches. If you look at 2020 matches, that error should not be there.

calebeby:
> @calebeby There's still a red-flag popup that says "no schema found". I tried to fix it the way we did for the offline-match-pages issue, but that didn't solve it. Any ideas?

I'm realizing that I didn't quite answer the question you asked. It looks like this error is coming from the `eventStats` variable. You should be able to add the handling to skip fetching there if no schema is found.

calebeby:
On small screens there is no space above the text:

screenshot

This is because the analysis page CSS removes the padding on small screens to save space and make the analysis table more full-screen.

calebeby:
Could we display a list of teams when there is no schema?
Also, when there is a schema but before the event starts, I think we should display the analysis table full of ?'s (I think that is the current behavior on the `dev` branch).

varCepheid:
My no-schema error fix doesn't actually work, it results in the Loader a lot and always says "404 page not found". I don't have time to fix it now.

varCepheid:
It's very difficult to figure out where the error is and to eliminate it in all cases without getting loaders on pages that should be able to load. I did my best, but I'm still getting a "no schema found" popup on this year's events and the layout doesn't look good on some others.

calebeby:
@varCepheid as you mentioned there are still some issues with the functionality and visual design. I'm not sure this issue is the most important/best use of your time since it will disappear (for this year) once we add a schema. Feel free to disagree with me though, if you think this issue is important let me know and I'll do a closer review with more feedback and maybe we can do another slack call to work out the remaining issues.

varCepheid:
You're right, I won't try to work on it.

PR #1498, opened 2022-04-08
Fixing "next match" feature

### LINKED ISSUE #1493: "Next match" feature isn't always accurate
Sometimes, it gives the match currently in progress.

varCepheid:
closes #1493 
At the top of the event and event-team pages, instead of one "next match," the page shows the three most relevant upcoming matches.


calebeby:
Hey @varCepheid, usually the scheduled match time and the predicted match time are wrong by 5-10 minutes. Do you think that instead of this solution we could show "upcoming matches" as 2-3 matches?

ev118:
I think the problem has been solved, though a bit of tweaking may be useful so that it looks neat (the Upcoming Matches header is off-center, even though it says justify-self: center; and the search bar sits right on top of Qual 1).

calebeby:
Hey @ev118, thanks for updating this! I think there are several opportunities to simplify the approach here, would you be open to scheduling a meeting this weekend to work through that with me?

calebeby:
@varCepheid can you delete commented-out code from this PR? Generally it is good practice to try to avoid committing commented-out code to dev/main branches.

varCepheid:
> Can you delete commented-out code from this PR? Generally it is good practice to try to avoid committing commented-out code to dev/main branches.
@calebeby Done.

src/utils/upcoming-matches.ts
```diff
@@ -0,0 +1,19 @@
...
+  for (const match of matches) {
+    if (
+      match.time &&
+      match.time.getTime() < currentTime + 1000 * 60 * 15 &&
+      match.time.getTime() > currentTime - 1000 * 60 * 4
+    ) {
+      retArr.push(match)
+    }
+  }
+  return retArr
```
calebeby, line 18: @varCepheid can you try using the `.filter` method for arrays here? It simplifies the code a little bit and makes it easier to see what is going on:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

It would look something like:


```js
return matches.filter((match) => match.time && match.time.getTime......)
```

Then you don't have to deal with creating the return array yourself and pushing items to it, the `filter` method is made to do exactly that automatically!

src/utils/upcoming-matches.ts
```diff
@@ -0,0 +1,19 @@
+import { ProcessedMatchInfo } from '@/api/match-info'
+// import { useCurrentTime } from './use-current-time'
```
calebeby, line 2: Can we remove this commented line of code?

src/routes/event-team.tsx
```diff
@@ -195,11 +184,36 @@ const EventTeam = ({ eventKey, teamNum }: Props) => {
...
+        <p
+          class={css`
+            color: gray;
+            text-align: center;
+            margin: 1rem 0 0.8rem;
+          `}
+        >
+          No Upcoming Matches
+        </p>
       )}
```
calebeby, line 217: @varCepheid I am wondering how helpful it is to have this "no upcoming matches" text:

image

What do you think about just leaving that part empty if there aren't any coming up?

src/utils/upcoming-matches.ts
```diff
@@ -0,0 +1,19 @@
...
+export const getUpcomingMatches = (
+  matches: ProcessedMatchInfo[],
+  currentTime: number,
+) => {
+  const retArr = []
+  for (const match of matches) {
+    if (
+      match.time &&
+      match.time.getTime() < currentTime + 1000 * 60 * 15 &&
+      match.time.getTime() > currentTime - 1000 * 60 * 4
```
calebeby, line 13: @varCepheid It looks like this only shows matches between 4 minutes ago and 15 minutes from now.

I am looking at the madtown throwdown event that is currently happening. Here's what it looks like on `main`:

https://peregrinefrc.com/events/2023mttd/teams/1678

image

This is helpful to me because it shows when the team will compete next, even though it is more than 15 minutes from now.

On this branch, here is what it shows:

image

The "no upcoming matches" text is a little confusing to me since it would make me assume that they do not have any matches coming up, while in reality they are probably queueing for the match (the time of the screenshot is 9:06 AM)

Another potential problem/improvement that we could implement is that matches that already have scores do not need to be shown in "upcoming matches" even if they were less than 4 minutes ago (since we know the match already finished). You can copy this logic from the next-incomplete-match file that was deleted.

Can I propose this logic for getUpcomingMatches instead?

```js
const ONE_DAY = 24 * 60 * 60 * 1000;
return matches
  // Only include matches that haven't finished yet (scores haven't been posted yet)
  .filter(match => match.redScore === undefined)
  // If 24h has passed since a match's predicted time without scores being posted,
  // it probably won't ever be played.
  // For example: tiebreaker matches that won't get played
  .filter(match.time.getTime() > currentTime - ONE_DAY)
  // limit to showing 3 matches
  .slice(0, 3)
```

This way, we can avoid all timing-based logic, and that avoids a bunch of time-related issues like field delays right before a match, and it still shows matches for the next day that teams will play in, which is nice.

It also would avoid showing for too long tiebreaker matches that never will be played, like https://www.thebluealliance.com/match/2023brd_f1m3 (since the red alliance won F1 and F2, the F3 tiebreaker does not need to be played, but the FMS never went and deleted that match)

src/routes/event-team.tsx
```diff
@@ -195,11 +184,36 @@ const EventTeam = ({ eventKey, teamNum }: Props) => {
       back={`/events/${eventKey}`}
       class={eventTeamStyle}
     >
-      {nextMatch && (
+      {upcomingMatches && upcomingMatches.length > 0 ? (
         <>
-          <h2 class={sectionStyle}>Next Match</h2>
-          <MatchDetailsCard match={nextMatch} eventKey={eventKey} link />
+          <Heading level={2}>Upcoming Matches</Heading>
```
calebeby, line 189: This is a little nitpicky, but the space between the "upcoming matches" heading and the actual match cards is half the size of the space between the other elements on the page.

Before this PR:

image

After this PR:

image

The new heading looks nice in terms of the underline and text styling, but it feels a little bit cramped without the even space around it.

calebeby:
@varCepheid I also noticed that the issue https://github.com/Pigmice2733/peregrine-frontend/issues/1493 is kind of vague and doesn't mention whether it is talking about the next match feature on the team@event page or on the event page. This PR only fixes the team@event page, so maybe we should leave that issue open when this is merged and open another PR for the event page.

varCepheid:
I agree that the issue is vague. I added a description on the PR, which is hopefully more clear. And the changes here do affect the event page; if it wasn't doing anything different, we should definitely fix that here.

varCepheid:
When I merged from `dev`, it brought a bunch of changes to `package-lock.json`. I tried to deal with all of them, but the file is so large that it was lagging a lot, and I think I made a mistake somewhere. Should I try to revert the commit?

calebeby:
@varCepheid sorry for not getting back to you sooner. You shouldn't ever need to manually merge a `package-lock.json` file. You can always just checkout the `package-lock.json` from either side of the merge, and then run `npm i` to make npm fix any issues or discrepancies from the `package.json` file. In this case, there are no changes to `package.json` so we don't even need to do that. Since the changes to `package-lock.json` aren't relevant to this PR, I went ahead and reverted them. (`git checkout origin/dev -- package-lock.json` is the easiest way to do this for a single file without necessarily reverting an entire commit, it just means "set this file to as it was on _ branch").

varCepheid:
Is there anything else that you need to review, or is this ready to push?

src/utils/move-focus-inside/index.ts
```diff
@@ -4,6 +4,7 @@ const tababblesStr = tabbables.join(',')
 
 export const moveFocusInside = (el: HTMLElement) => {
   if (el.matches(tababblesStr)) return el.focus()
+  // eslint-disable-next-line no-warning-comments
```
calebeby, line 7: This appears to be unrelated to this PR

src/utils/clean-field-name.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-warning-comments */
```
calebeby, line 1: This appears to be unrelated to this PR

calebeby (approved):
Looks good just left some comments about the eslint-disable comments

calebeby (approved):

PR #1500, opened 2022-04-11
changed saved match location to a preact-state

varCepheid:


calebeby:
@varCepheid do you mind writing a PR description to explain what this does and why?

calebeby:
Thanks for fixing this @varCepheid! I think the root cause of this bug is that the URL manager (which handles URL changes and passing the event key and match key to the ScoutPage component) doesn't know that the URL changes when we call history.replaceState. It turns out, we have a small wrapper function for history.replaceState called `updateUrl` which when it is called will in turn call history.replaceState, and handle updating the URL-based props including event key and match key.

Can we use that instead of storing this value in state?

I think it would work like this:

```ts
onMatchSelect={(newMatchKey) =>
  updateUrl(
    `/events/${eventKey}/matches/${newMatchKey}/scout`,
    true,
  )
}
```

varCepheid:
Made that change and it still works. Can I merge?

src/utils/compare-events/index.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable caleb/jest/expect-expect */
```
calebeby, line 1: I think you can undo this change? Not sure what it's for 😕

varCepheid: There were a lot of issues being thrown about tests that didn't work for some reason. I can take it out if you don't want it, but I think it'll make the lint cleaner in the future.

calebeby: Oh I see what this is flagging. Can you make a separate PR to add it to the project eslint config instead of just this individual file? https://github.com/Pigmice2733/peregrine-frontend/blob/dev/package.json#L118

calebeby (approved):

PR #1504, opened 2022-06-03
Don't have to "override" an edit of a report

### LINKED ISSUE #1496: overriding reports on edit
When I save a report edit, it shows a pop-up that asks if I want to override the report I had submitted previously. It should override automatically since that's what an edit is for.

varCepheid:
closes #1496


varCepheid:
The issue I'm having is that when I do edit the report, it doesn't save to the server even though I have a good connection. My locally saved reports won't sync, either.

calebeby:
We fixed this with a backend change

PR #1509, opened 2022-06-07
a better 404 page

### LINKED ISSUE #1505: 404 page is quite bland
Fix the 404 page so that it looks better than just a block of text that says '404;' at least it should include a button/link back to an existent page on the site.

ev118:
closes #1505


varCepheid:
We're waiting on this until match-page-errors is pushed so that you can use the card component I'm making.

PR #1513, opened 2022-06-09
404 error page

### LINKED ISSUE #1505: 404 page is quite bland

ev118:
closes #1505 
The most recent commit used the new component pulled in from match-page-errors but didn't change much else. The aesthetics of the card still need to be worked on but it would be beneficial to do that once the changes from match-page-errors and event-page-errors have also been merged.


src/routes/404-error.tsx
```diff
@@ -0,0 +1,36 @@
...
+      class={css`
+        font-size: 30pt;
+        font-weight: bold;
+      `}
+    >
+      404
+    </Heading>
+    <ErrorCard
+      errorText={
+        'The requested URL was not found. Use the button below to go to the home page.'
```
varCepheid, line 28: move the second sentence to errorText2 so that there isn't a one-word line

varCepheid:
@calebeby We want your opinion on an appearance. Line 21 in `routes/404-error` sets the font family to monospace. I prefer the default font family, which is sans serif. Which do you think looks better?

calebeby:
@varCepheid Usually I prefer sans-serif unless the UI is showing code.

PR #1514, opened 2022-06-10
Fixed errors and aesthetics in report-overriding message

varCepheid:
closes #1508
closes #1507

PR #1515, opened 2022-06-10
Fix errors in the dialog to override a report

### LINKED ISSUE #1507: change text formatting in popup for overriding report conflicts

### LINKED ISSUE #1508: automatically overrides on new conflicting report
When a new report is created with the same team, match, and reporter as another that already exists, the new one automatically replaces the old one with no confirmation. A dialog should appear warning the user that they are overriding a pre-existing report, as it does when a report is edited to conflict with another.

varCepheid:
closes #1507 
closes #1508 

Makes the dialog pop up if there is an actual conflict, but not if it's just an edit.
Makes the description text in a dialog thicker so it's visible.
Adds a dialog on delete (it's not perfect but it works).

### DIFF
--- a/src/components/dialog.tsx
+++ b/src/components/dialog.tsx
@@ -100,7 +100,13 @@ export const DialogDisplayer = () => {
       <div tabIndex={0} class={dialogStyle} aria-modal="true" role="dialog">
         {dialog.title && <h1>{dialog.title}</h1>}
         {typeof dialog.description === 'string' ? (
-          <p>{dialog.description}</p>
+          <p
+            class={css`
+              font-weight: 500;
+            `}
+          >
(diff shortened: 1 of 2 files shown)


varCepheid:
This replaces PR #1514, which I accidentally merged handle-network-errors into. It has all the same changes I made.

calebeby:
@varCepheid Can you add screenshots and testing instructions to this PR?

varCepheid:
I think they're probably in #1514 because that was the original PR. I'll move them in here when I get the chance.

varCepheid:
I added some testing instructions in the PR description. Does that help? I checked that the changes I made did what I want them to using the deploy preview.

PR #1516, opened 2022-06-10
Handle network errors

varCepheid:


calebeby:
@varCepheid this is an intimidatingly large PR to review and test. Is it OK with you if I hold off on reviewing it until all the other open PRs get merged? That way we can avoid conflicts (even if they are not explicitly flagged by git) and make sure that everything gets tested in-depth.

varCepheid:
Yes, that's fine.

PR #1520, opened 2023-01-13
updates to leaderboard: years dropdown and card links

### LINKED ISSUE #1518: sort years correctly on leaderboard
image
Year list should be sorted newest to oldest.

varCepheid:
closes #1518


varCepheid:
Accidentally pushed a change for a different issue to this, but they're both leaderboard-related, so it's fine.

package.json
```diff
@@ -142,7 +142,7 @@
           ]
         }
       ],
-      "no-warning-comments": "off"
```
calebeby, line 145: @varCepheid I noticed you have changed this in a bunch of PR's and I was wondering why? What is the intended value, and can we keep it the same across all files?

varCepheid: I don't know why I've changed it in multiple branches. The purpose is to stop making it throw errors or warnings for the TODO comments in a couple files. I can change it to "warn" if you like, but "error" seems to be the wrong approach when the TODO comments seem important.

calebeby (approved):

PR #1521, opened 2023-01-13
frontend page for creating a new realm

### LINKED ISSUE #1510: create realms from frontend
will require backend editing as well

varCepheid:
closes #1510

### DIFF
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -44,6 +44,10 @@ const routes = [
     component: () => import('./routes/login'),
   },
   {
+    path: '/newrealm',
+    component: () => import('./routes/new-realm'),
+  },
+  {
     path: '/leaderboard',
     component: () => import('./routes/leaderboard'),
   },
--- a/src/routes/new-realm.tsx
+++ b/src/routes/new-realm.tsx
@@ -1,44 +1,3 @@
-import TextInput from '@/components/text-input'
-import Card from '@/components/card'
-import Page from '@/components/page'
-import { createUser } from '@/api/user/create-user'
-import Button from '@/components/button'
-import {
-  minUsernameLength,
-  maxUsernameLength,
-  minPasswordLength,
-  maxPasswordLength,
-} from '@/constants'
-import { useState } from 'preact/hooks'
-import { css } from 'linaria'
-import { Form } from '@/components/form'
-import { Dropdown } from '@/components/dropdown'
-import { getRealms } from '@/api/realm/get-realms'
-import { usePromise } from '@/utils/use-promise'
-import { ErrorBoundary, useErrorEmitter } from '@/components/error-boundary'
-import { authenticate } from '@/api/authenticate'
-import { route } from '@/router'
-import { Realm } from '@/api/realm'
-
-const signUpStyle = css`
-  padding: 1.5rem;
-`
-
-const cardStyle = css`
-  padding: 1.5rem 2rem;
-  width: 20rem;
-  margin-left: auto;
-  margin-right: auto;
-  & > * {
-    margin-left: 0;
-    margin-right: 0;
-  }
-`
-
-const dropdownClass = css`
-  padding: 0.4rem;
-`
-
 const SignUpForm = () => {
   const [isLoading, setIsLoading] = useState(false)
   const [firstName, setFirstName] = useState('')
@@ -100,9 +59,6 @@ const SignUpForm = () => {
             getKey={(v) => v.id}
             getText={(v) => v.name}
           />
-          <div>
-            <Button disabled={isLoading || !isValid} href={'/newrealm'}>New Realm</Button>
-          </div>
           <Button disabled={isLoading || !isValid}>
             {isLoading ? 'Signing Up' : 'Sign Up'}
           </Button>


varCepheid:
At line 37 in `new-realm`, the server gets a request to POST a new realm, even if that realm is already in the database. Is there a check that should happen in the frontend code to make sure that doesn't happen, or will the backend take care of it?

varCepheid:
Also, I'm going to leave this as draft until the Vite stuff takes effect and I can use a localhost to debug.

varCepheid:
I don't totally understand what the test files are doing. I messed around with it a little bit, but I couldn't get the check to pass. I only changed the two under components. Can you take a look at those and either recommend something to do or fix it yourself?

calebeby:
@varCepheid The test fails because the login form is broken. The test is intended to catch bugs like this, so it is doing its job perfectly!

If you go to the preview for this PR: https://deploy-preview-1521--peregrine.netlify.app/leaderboard

Once you type a valid username and password, the "submit" button should stop being disabled, so you can click it. (That's what happens on https://dev.peregrinefrc.com/leaderboard)

I am not sure exactly what you changed in your PR that caused this behavior to break, but a good starting point would be to try undoing each of your changed files one by one until it is fixed, so that you can narrow down what caused the problem.

calebeby:
@varCepheid I am guessing that this change is what's breaking it: https://github.com/Pigmice2733/peregrine-frontend/pull/1521/files#diff-7f29dc5f7895e5965f3a69c6f27d41be530b11108dba3c035c8287ec491b4cf2R22

I am not sure the intended behavior of that change, but because of the change, the ref never gets passed in to the form element, so the ref never gets set, so there is no way for the component to check whether its form is valid.

calebeby:
Also @varCepheid once you fix the bug, if you could undo the unrelated parts of this PR that'd be great! Smaller PR's are much easier to review and merge, and they'll make it easier to track down changes in the git history once we merge them too!

varCepheid:
A lot of the changes that aren't connected to this issue are fixing errors that make the checks fail. Is it better to fix them in separate PRs and put this one on hold?

calebeby:
@varCepheid The checks are passing on the `dev` branch, and if this branch is up to date with that branch, then the checks should be passing here too. I think at least some of the changes in this PR aren't needed to make checks pass.

src/routes/signup.tsx
```diff
@@ -90,16 +103,25 @@ const SignUpForm = () => {
...
+              class={dropdownClass}
+              options={realms}
+              required
+              onChange={(v) => setRealmId(v.id)}
+              getKey={(v) => v.id}
+              getText={(v) => v.name}
+            />
+            <div class={textStyle}>
+              Don&apos;t see your team? Ask a captain or mentor to make a new
+              realm.{' '}
```
calebeby, line 119:
```suggestion
              {"Don't see your team? Ask a captain or mentor to make a new realm."}
```
Prettier might format this a little different. Not sure if it will change it to multiple lines.

src/routes.ts
```diff
@@ -43,6 +43,10 @@ const routes = [
     path: '/login',
     component: () => import('./routes/login'),
   },
+  {
+    path: '/newrealm',
```
calebeby, line 47: Nit: existing URLs with multiple "words" use dashes (e.g. `saved-reports`)

```suggestion
    path: '/new-realm',
```

src/routes/signup.tsx
```diff
@@ -90,16 +103,25 @@ const SignUpForm = () => {
...
+            </div>
+            <Button disabled={isLoading} href={'/newrealm'}>
+              New Realm
+            </Button>
```
calebeby, line 123: To me, this "new realm" button comes across as way too visually prominent. I could totally see someone filling out this signup form and clicking the big "new realm" button accidentally instead of clicking "sign up" as intended.

Also, I think we can un-bold the "don't see your team..." text.

I think making the new realm be a link in the text would reduce its visual prominence, making it more clear that "sign up" is the primary action for this page. Here's what that would look like:

image

calebeby:
@varCepheid I have a question about how this flow works for users. If a team was to create a new realm, now that realm has no users. So in order for them to start making approved users for their realm, they'd need to make an unverified user in that realm, and still go through the manual process of contacting pigmice and getting their account verified and set to admin. Then they could go and make other users.

Is it intended to still have that manual step there or were you envisioning something different?

calebeby (changes requested):

src/routes/signup.tsx
calebeby: Actually, if I remember right, you might not even need the curly brackets and quotes. I think you can just put an apostrophe in and it will convert to JS correctly and the HTML escaping will happen automatically

calebeby:
@varCepheid it looks like newly-created realms have `shareReports` set to `false`, while all the previous realms have it set to `true`. Is that intentional?

varCepheid:
The `shareReports` bit was unintentional and I fixed it. My solution to the no-admins problem was to make the new-realm form have all the same fields as the signup form, so it creates the first account in the same action in the new realm and sets it to be an admin.

src/routes/signup.tsx
```diff
@@ -90,16 +102,23 @@ const SignUpForm = () => {
...
+              emptyLabel="Select a realm"
+              class={dropdownClass}
+              options={realms}
+              required
+              onChange={(v) => setRealmId(v.id)}
+              getKey={(v) => v.id}
+              getText={(v) => v.name}
+            />
+            <div class={textStyle}>
+              {"Don't see your team? Ask a captain or mentor to "}{' '}
```
calebeby, line 117: There is an extra space that gets collapsed here.

```suggestion
              {"Don't see your team? Ask a captain or mentor to "}
```

varCepheid:
I realized that making a user automatically an admin A. is a bad idea and B. gives us no oversight of new realms. I think we should just indicate that a new user in a new realm needs to join the Slack to become an admin.

calebeby:
@varCepheid Good point. However, if the backend already allows it, then people still could _technically_ do it themselves (not a big fan of "security by obscurity"). Maybe we should make the backend have a constraint on creating new admin users/realms.

varCepheid:
There actually is already a backend constraint, which meant the user creation request wasn't going through, because a non-admin user (in this case not even logged in) can't create users. I think the ideal way to do this is to create only the realm and then force the would-be admin to communicate with us about creating their user. The alternative is to create a dummy super-admin and log in as that user for only the time it takes to create the new user.

varCepheid:
After talking with Elijah, I realized that of course a non-user can create a user because the signup page does it. The actual problem was in the return type of the `createRealm` method, but that's largely unimportant. Currently, the new-realm form creates the realm with an unverified user in it and directs the user to email me and get connected to the Slack. From there, we can verify them and make them an admin. I think this system puts the least burden on us but still allows us oversight of who is running realms.

PR #1522, opened 2023-01-15
change back button on match page to last viewed page

### LINKED ISSUE #1497: back button from match page
"On mobile, when we click on a match from a team’s page, we will get to the match page normally. However, when clicking the back button, it takes us back to the event page of the app rather than the page for that team."

varCepheid:
closes #1497

### DIFF
--- a/src/routes/event-match.tsx
+++ b/src/routes/event-match.tsx
@@ -163,7 +163,7 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
   return (
     // page setup
     <Page
-      back={`/events/${eventKey}`}
+      back={() => window.history.back()}
       name={
         m.group +
         (m.num ? ' Match ' + m.num : '') +


calebeby:
@varCepheid The way I designed the "back system" is that the back button goes "up a level", when that makes sense, and when there is no logical "up a level", it will perform the same action as the browser back button (go to the previous page).

Sometimes, that system is confusing. Instead of changing the behavior on this one individual page, can you propose a new logical system for the back button across all pages?

calebeby:
I acknowledge that the current system is confusing. Do you think it'd make sense for the back button in _all_ cases to work the same as the browser back button? I would not be opposed to that kind of change.

However, if we did that, then there would be no easy way to, for example, navigate from a match page to the event corresponding to that match, unless that is the page you came from.

What do you think would make the most sense?

varCepheid:
I think your fix makes more sense, and there should be a link to the event page on the match page.

varCepheid:
Note to self: finish working on match details card, seems to only be updating on event page; then fix the type errors.

varCepheid:
@calebeby routes/report, line 51: The back button had stopped editing mode when the report was being edited without changing the page. Now, the back button will always go to the browser's back no matter what. Should there be a button on the report page to stop editing, or is the Save Report button at the bottom good enough?

calebeby:
@varCepheid Can you make it so that when you click edit it goes to a different URL? e.g. when you are on https://peregrinefrc.com/reports/1749 and you click "edit", can it navigate to https://peregrinefrc.com/reports/1749/edit, so that when you click the back button from the edit page it goes back to the view report page?

src/components/report-editor.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable caleb/@typescript-eslint/restrict-plus-operands */
```
calebeby, line 1: @varCepheid what is this change for?

src/components/match-card.tsx
```diff
@@ -111,11 +136,19 @@ export const MatchDetailsCard = memo(
...
+              {' '}
+              at <a href={`/events/${eventKey}`}>{eventName}</a>{' '}
+            </div>
+          )}
+          {match.time && (
+            <time dateTime={match.time.toISOString()} class={dateTimeStyle}>
+              {formatTime(match.time)}
+            </time>
+          )}
+        </div>
```
calebeby, line 151: @varCepheid When I look at a match page, this placement of the link to the event page isn't really intuitive to me:

https://deploy-preview-1522--peregrine.netlify.app/events/2023orwil/matches/qm1

To me it feels like the link would be better suited in the navigation bar instead of in the match information card (especially to keep the match information card more consistent between pages):

2023-06-10T17:03:55,740243471-07:00

calebeby: Having the link in the navigation bar would also be something we can do across pages - for example this page (team@event) should probably have a link to the event in it (which was previously a back button). It could have the event name in the navigation bar link to the event:

https://deploy-preview-1522--peregrine.netlify.app/events/2023orwil/teams/1425?stat=auto::Docked

Same for team-matches@event: https://deploy-preview-1522--peregrine.netlify.app/events/2023orwil/teams/1425/matches

and team-comments@event: https://deploy-preview-1522--peregrine.netlify.app/events/2023orwil/teams/1425/comments

I'm sure there are lots of others I'm not thinking of. I think for any event-specific page, there should be a way to get from that page to the event page.

varCepheid: I did this and it works well by what I can see. In addition to the pages you mentioned, I added it to the analysis table page.

src/components/match-card.tsx
```diff
@@ -40,6 +40,7 @@ const matchCardStyle = css`
 const matchTitleStyle = css`
```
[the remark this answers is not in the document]
calebeby: Looking a little more closely at some pages, I'm not sure if these changes were intentional.

Here's how the match list looks on dev:

image

And on this branch:

image

src/routes/event-team.tsx
```diff
@@ -156,12 +157,22 @@ const teamHeadingSpanStyle = css`
   margin-right: 0.3rem;
 `
 
+const headerLinkStyle = css`
+  color: white;
+
+  &:hover {
+    color: ${greenOnPurple};
```
calebeby, line 164: @varCepheid I have a hard time with the green text on purple background. To me, it doesn't seem to match very well. Also, there is not very good contrast between the hover text color and the background color, which is an accessibility concern. https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast

I tried out some other colors to see what might look a little better and be more legible in terms of contrast. Light blue seems a little better to me, but it still didn't look quite right to me. Ultimately I think that white text on purple just looks a lot better. I don't think that the hover style in this case is really necessary, so I'd be happy with just letting the text remain white on hover. If having some sort of hover effect is important to you, maybe can we try setting `text-decoration-color` instead? It would be a little more subtle and wouldn't hurt contrast.

calebeby (changes requested):
Hi @varCepheid! Thanks for your patience with me! I left some feedback. I got a chance to review almost all of the files in this PR, except I haven't had time to closely review the changes to the report editor yet (when big blocks of code are moved the diff view doesn't really show what, if anything was actually changed, so it takes longer to compare)

src/components/match-card.tsx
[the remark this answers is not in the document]
varCepheid: I think I did that when I was trying to fit the link into the card. None of it seems to be doing anything important.

varCepheid:
Thanks for all the notes! I went with a `text-decoration-color` attribute for the last point and I think it looks good. I won't merge until you have a chance to look at the `report-editor` component.

varCepheid:
@calebeby I've addressed all of the changes you've requested. Can you add another review or approve the changes?

src/colors.ts
```diff
@@ -5,7 +5,7 @@ export const lightGrey = '#E8E8E8'
 export const faintGrey = '#f5f5f5'
 export const blue = '#295dc0'
 export const red = '#c30000'
-export const greenOnPurple = '#4FAC40'
-export const redOnPurple = '#D61E42'
+export const greenOnPurple = '#47D232'
+export const redOnPurple = '#FF3C2E'
```
calebeby, line 9: @varCepheid These changes impact the boolean stats display on the team@event page.

On this branch:

https://deploy-preview-1522--peregrine.netlify.app/events/2023wamc/teams/360

image

On dev:

https://dev.peregrinefrc.com/events/2023wamc/teams/360

image

I chose the colors that are currently on dev because they maintain enough contrast against the white check mark and x. The colors that you changed them to have a lot less contrast against white, especially the bright green.

Also, it's not clear to me why you changed redOnPurple in this PR since you aren't using it in this PR.

src/components/page.tsx
```diff
@@ -46,19 +46,27 @@ const headerText = css`
...
     name: ComponentChildren
-    back: string | (() => void) | false
+    showBackButton?: boolean
     class?: string
     wrapperClass?: string
   }
 >
 
-const Header = ({ back, name }: Omit<Props, 'class'>) => {
+const Header = ({ name, showBackButton }: Omit<Props, 'class'>) => {
```
calebeby, line 69:
```suggestion
const Header = ({ name, showBackButton = true }: Omit<Props, 'class'>) => {
```

@varCepheid would it be more expressive to set the default value to true? Then people who read the function signature can see that the parameter defaults to true, and the code below that uses it could be simplified to be easier to read too.

src/routes/event-analysis.tsx
```diff
@@ -25,13 +25,18 @@ const teamStyle = css`
...
   const schema = useSchema(eventInfo?.schemaId)
 
   return (
     <Page
-      name={`Analysis - ${eventInfo ? eventInfo.name : eventKey}`}
-      back={`/events/${eventKey}`}
+      name={
+        <>
+          {'Analysis - '}
+          <a href={`/events/${eventKey}`}> {eventName} </a>
```
calebeby, line 37: We do not need these extra spaces
```suggestion
          <a href={`/events/${eventKey}`}>{eventName}</a>
```

src/routes/event-match.tsx
```diff
@@ -163,12 +163,11 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
     <Page
-      back={`/events/${eventKey}`}
       name={
-        m.group +
-        (m.num ? ' Match ' + m.num : '') +
-        ' - ' +
-        (event ? event.name : eventKey)
+        <>
+          {m.group + (m.num ? ' Match ' + m.num : '') + ' - '}
+          <a href={`/events/${eventKey}`}> {event ? event.name : eventKey} </a>
```
calebeby, line 169: We do not need these extra spaces

```suggestion
          <a href={`/events/${eventKey}`}>{event ? event.name : eventKey}</a>
```

Also, this can use optional chaining syntax:

```suggestion
          <a href={`/events/${eventKey}`}> {event?.name || eventKey} </a>
```

src/routes/event-match.tsx
```diff
@@ -163,12 +163,11 @@ const EventMatch = ({ eventKey, matchKey }: Props) => {
...
     // page setup
     <Page
-      back={`/events/${eventKey}`}
       name={
-        m.group +
-        (m.num ? ' Match ' + m.num : '') +
-        ' - ' +
-        (event ? event.name : eventKey)
+        <>
+          {m.group + (m.num ? ' Match ' + m.num : '') + ' - '}
```
calebeby, line 168: Can you use a template string (string with backticks) instead of `+`? Not sure why I used `+` when I wrote this, silly me :stuck_out_tongue_winking_eye:

src/routes/event-team-comments.tsx
```diff
@@ -46,8 +50,12 @@ const EventTeamComments = ({ eventKey, teamNum }: Props) => {
 
   return (
     <Page
-      name={`Comments: ${teamNum} @ ${eventInfo ? eventInfo.name : eventKey}`}
-      back={`/events/${eventKey}/teams/${teamNum}`}
+      name={
+        <>
+          {`Comments: ${teamNum} @ `}
+          <a href={`/events/${eventKey}`}> {eventName} </a>
```
calebeby, line 56: We do not need these extra spaces
```suggestion
          <a href={`/events/${eventKey}`}>{eventName}</a>
```

src/routes/event-team-matches.tsx
```diff
@@ -25,8 +25,12 @@ const EventTeamMatches = ({ eventKey, teamNum }: Props) => {
   const matches = useEventMatches(eventKey, 'frc' + teamNum)
   return (
     <Page
-      name={`Matches: ${teamNum} @ ${eventName}`}
-      back={`/events/${eventKey}/teams/${teamNum}`}
+      name={
+        <>
+          {`Matches: ${teamNum} @ `}
+          <a href={`/events/${eventKey}`}> {eventName} </a>
```
calebeby, line 31: We do not need these extra spaces
```suggestion
          <a href={`/events/${eventKey}`}>{eventName}</a>
```

src/routes/event-team.tsx
```diff
@@ -178,9 +180,10 @@ const EventTeam = ({ eventKey, teamNum }: Props) => {
     <Page
       name={
         <span class={pageHeadingStyle}>
-          <span class={teamHeadingSpanStyle}>{`${teamNum} @ ${
-            eventInfo ? eventInfo.name : eventKey
-          }`}</span>
+          <span class={teamHeadingSpanStyle}>
+            {`${teamNum} @ `}
+            <a href={`/events/${eventKey}`}> {eventName} </a>
```
calebeby, line 185:
```suggestion
            <a href={`/events/${eventKey}`}>{eventName}</a>
```

calebeby (changes requested):
Great work here! I still have yet to review the report editor changes, I can't remember why those are in this PR and I don't have time to look closely right now.

varCepheid:
The changes to the `report` and `saved-report` routes are because the logic on the back button used to discard changes to a report being edited, but with this PR in order to maintain that logic we needed new pages.
> We Do Not Need These Extra Spaces --Caleb Eby

src/components/report-viewer.tsx
```diff
@@ -130,7 +133,7 @@ export const ReportViewer = ({ report, onEditClick }: Props) => {
 
       {/* links to the author of the report */}
       <ProfileLink reporterId={reporterId} />
-      {onEditClick && <Button onClick={onEditClick}>Edit</Button>}
+      {onEditClick && <Button href={onEditClick}>Edit</Button>}
```
calebeby, line 136: Let's rename `onEditClick` to `reportEditorLink` since it is not an event handler anymore, but the name implies that it is.

src/routes/event.tsx
```diff
@@ -73,7 +69,6 @@ const Event = ({ eventKey }: Props) => {
         </Heading>
         {newestIncompleteMatch && (
           <MatchDetailsCard
-            key={newestIncompleteMatch.key}
```
calebeby, line 76: I am not sure why you removed this attribute. Having a key attribute helps Preact with diffing, so that it knows the difference between when items were changed, added, reordered, or removed. I think there was some weird edge case where when this page rerendered (due to the newestIncompleteMatch switching), it might have reset the input on the event matches list, or something like that.

src/routes/report-edit.tsx
```diff
@@ -0,0 +1,84 @@
...
+  onSaveLocally: (report: OfflineReport) => void
+  onDelete: () => void
+}) => {
+  return (
+    <Page name="Edit Report" class={reportPageStyle}>
+      <ReportEditor
+        initialReport={report}
+        onSaveSuccess={(report) => {
+          onSaveSuccess(report)
+        }}
```
calebeby, line 34: This can be shortened to `onSaveSuccess={onSaveSuccess}` (same below)

src/routes/saved-report.tsx
```diff
@@ -1,60 +1,66 @@
...
+        <ReportViewer
+          report={report}
+          reportEditorLink={
+            canEdit ? `/saved-reports/${report.key}/edit` : undefined
+          }
+        />
+      </Card>
+    </Page>
+  )
+}
```
calebeby, line 49: Before this PR, `ReportPage` was shared between the offline report editor and the "online" report editor. It looks like you split it out here. My concern is that the two pages (which are intended to look and work essentially the same) may now get out of sync in case someone forgets to edit both. For example, someone may make a CSS change in one but not the other. The only real difference between the two is how they handle fetching/storing the report data. Is there a way to combine them back together?

calebeby (commented):
@varCepheid This is looking really good! I finally remembered to come back here to look at the report viewer/editor changes. I left some comments.
Also feel free to bug me on slack if I forget to review things again!

src/routes/event.tsx
varCepheid: I removed it because the `MatchDetailsCard` component doesn't have `key` as a property. I don't know what it's passing into.

src/routes/saved-report.tsx
varCepheid: I did what I thought would work well here, but I didn't do a thorough comparison of every pair of files among the six that existed. There isn't enough overlap between the viewer and the editor to consolidate.

PR #1523, opened 2023-01-15
created an alert when local reports are uploaded

### LINKED ISSUE #1100: Add an alert for when you reconnect and offline reports get uploaded
When you reconnect and the offline reports are able to be uploaded, it should make an alert that says something like "3 offline reports were uploaded"

Here is the code that uploads offline reports: https://github.com/Pigmice2733/peregrine-frontend/blob/dev/src/api/report/submit-report.ts#L34

varCepheid:
closes #1100

### DIFF
--- a/src/api/report/submit-report.ts
+++ b/src/api/report/submit-report.ts
@@ -2,6 +2,8 @@ import { request } from '../base'
 import { Report, OfflineReport } from '.'
 import { useEffect, useState } from 'preact/hooks'
 import { CancellablePromise } from '@/utils/cancellable-promise'
+import { createAlert } from '@/router'
+import { AlertType } from '@/components/alert'
 
 /**
  * Uploads a report via PUT or POST, depending on if the report already has an ID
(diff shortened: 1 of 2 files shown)


src/api/report/submit-report.ts
```diff
@@ -53,6 +55,12 @@ export const uploadSavedReports = async () => {
     [] as Report[] | Promise<Report[]>,
   )
   localStorage.setItem(SAVED_REPORTS, JSON.stringify(unsuccessfulReports))
+
+  const uploadedReports = savedReports.length - unsuccessfulReports.length
+  createAlert({
+    type: AlertType.Success,
+    message: `${uploadedReports} reports were uploaded!`,
+  })
```
calebeby, line 63: @varCepheid the `uploadSavedReports` function runs 2 seconds after the initial load, and then every 30 seconds. Since there isn't a check here of uploadedReports !== 0, this ends up happening (a new one every 30 seconds):

image

The [deploy preview](https://deploy-preview-1523--peregrine.netlify.app/) that is automatically created for each PR is super helpful in testing out your changes! Peregrine has a lot of moving pieces and it is important to carefully test all code changes to make sure everything works as expected.

calebeby (changes requested):

varCepheid:
I notice it creates an additional dialog when a report is saved locally, saying "NetworkError when attempting to fetch resource" in a red alert. I think the event-match page isn't able to load something, but I'm not sure what.

PR #1524, opened 2023-01-15
fixed the lint errors

varCepheid:


ev118 (approved):

src/components/text-input.tsx
```diff
@@ -22,7 +22,7 @@ const innerClass = css`
     outline: none;
   }
 
-  &${'.' + hasFocusedClass}:invalid {
+  &${clsx('.', hasFocusedClass)}:invalid {
```
calebeby, line 25: @varCepheid I think this will break the functionality. `clsx` will add a space in between its parameters so the output will look like `. something:invalid`, which is an invalid selector. The original code produces output like `.something:invalid`, which is a valid selector.

May I suggest changing the code like this?

```
  &.${hasFocusedClass}:invalid {
```

PR #1530, opened 2023-03-11
add dialogs when admins edit other users

### LINKED ISSUE #1094: Add alerts when editing users in the admin pages

varCepheid:
closes #1094
When an admin or super-admin edits the name, username, or password of another user, the app should create a dialog to confirm the change. The dialog shouldn't be created if the user is editing their own profile.

### DIFF
--- a/src/routes/user.tsx
+++ b/src/routes/user.tsx
@@ -173,9 +173,11 @@ const DeleteUserButton = ({ user }: { user: UserInfo }) => {
 const SetPasswordButton = ({
   user,
   self,
+  refetch,
 }: {
   user: UserInfo
   self: boolean
+  refetch: () => Promise<void>
 }) => {
   const emitError = useErrorEmitter()
   const alertOnSaveAdmin = async () => {
@@ -202,6 +204,7 @@ const SetPasswordButton = ({
           modifyUser(user.id, { password }).catch(emitError)
         }
       }}
+      refetch={refetch}
     >
       {(_password, _icon, startEditing) => (
         <Button flat onClick={startEditing}>
@@ -223,6 +226,7 @@ interface EditableTextProps {
   editable?: boolean
   label: string
   value: string
+  refetch: () => Promise<void>
 }
 
 const editableTextStyle = css`
@@ -238,6 +242,7 @@ const EditableText = ({
   label,
   editable,
   value: originalValue,
+  refetch,
   ...rest
 }: EditableTextProps & TextInputProps) => {
   const [isEditing, setIsEditing] = useState(false)
@@ -248,6 +253,7 @@ const EditableText = ({
   const closeEditor = () => {
     setIsEditing(false)
     setIsSaving(false)
+    refetch()
   }
   const openEditor = () => {
     setIsEditing(true)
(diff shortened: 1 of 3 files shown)


varCepheid:
The refetch function still isn't doing what it's supposed to, but I'll make that a separate PR. This solves the one change.

src/routes/user.tsx
```diff
@@ -381,7 +425,13 @@ const UserProfileCard = ({
           />
         </dl>
         {editable && !isCurrentUser && <DeleteUserButton user={user} />}
-        {editable && <SetPasswordButton user={user} />}
+        {editable && (
+          <SetPasswordButton
+            user={user}
+            self={isCurrentUser}
```
calebeby, line 431: @varCepheid Do you want to rename the prop to `isCurrentUser`? I think that name is more clear than `self`, which makes me think of [other things](https://en.wikipedia.org/wiki/This_(computer_programming))

src/routes/user.tsx
```diff
@@ -170,8 +170,26 @@ const DeleteUserButton = ({ user }: { user: UserInfo }) => {
...
+  user: UserInfo
+  self: boolean
+  refetch: () => Promise<void>
+}) => {
   const emitError = useErrorEmitter()
+  const alertOnSaveAdmin = async () => {
+    if (self) return
+    const confirmation = await createDialog({
+      title: 'Confirm Edit',
+      description: 'Are you sure you want to make this change?',
```
calebeby, line 187: @varCepheid Can you tell me more about why we are sending an alert when you edit another user, but not yourself?

varCepheid: That was my understanding of what we needed from this set of changes. If we want it to confirm for all users, I can change it.

calebeby: I think it would make more sense to me for it to either show the alert for all users or change the alert text to say something that makes it clear why it might be unexpected that you are editing another user (e.g. "Are you sure you want to edit user _____ (this is not your own account)" to prevent accidentally editing someone else when you meant to edit yourself).

varCepheid:
@calebeby I made the change you requested. It seems like you didn't mark your comments as a review, so I can't request a new review from you.

src/routes/user.tsx
```diff
@@ -170,7 +170,26 @@ const DeleteUserButton = ({ user }: { user: UserInfo }) => {
   )
 }
 
-const SetPasswordButton = ({ user }: { user: UserInfo }) => {
+const alertOnSaveAdmin = async (selfEdit: boolean, user: UserInfo) => {
```
calebeby, line 173: Can we rename `selfEdit` to `isCurrentUser` for consistency? (I think they mean the same thing but it is easier to understand if they match)

calebeby:
> The refetch function still isn't doing what it's supposed to, but I'll make that a separate PR. This solves the one change.

@varCepheid Do you want to remove the refetch-related changes from this PR since it looks like it still isn't working?

calebeby (commented):
@varCepheid This is excellent! Super impressed! I left two tiny bits of feedback, but this is just about ready to go!

calebeby (approved):
:shipit: Thank you!

PR #1533, opened 2023-08-02
about page

### LINKED ISSUE #1531: about page
Add a page that has information about Peregrine and how teams can start using it. This should be either a button in the menu or on the homepage. This could also include adding the buttons from the menu (leaderboard, user list, login page, profile) to the top of the homepage above the events list.

varCepheid:
closes #1531 
This PR creates an about page, linked from the sidebar menu, that has information about what Peregrine is, how teams can start using it or get help using it, and some useful links and contact information.


varCepheid:
I put in a version number, but I can easily take that out if we don't like it. Also, whoever reviews this, please proofread the text.

src/routes/about.tsx
```diff
@@ -0,0 +1,109 @@
...
+const AboutPage = () => {
+  return (
+    <Page name="About" back="/" class={pageStyle}>
+      <h1 class={headerStyle}>Welcome to Peregrine!</h1>
+      <div class={informationStyle}>
+        <p>
+          <span class={boldText}>
+            Peregrine is a completely free, open-source scouting app for FRC
+            competitions.
+          </span>{' '}
```
calebeby, line 49: @varCepheid Can we use semantic HTML elements for text like this instead of putting CSS classes on `<span>`s? [Here's an article explaining how semantic elements work and benefit people using assistive technology](https://blog.hubspot.com/website/semantic-html). Plus, I think we'll end up with less code to do the same thing!

In the specific case of boldText, we can change `<span class={boldText}>` → `<strong>`

src/routes/about.tsx
```diff
@@ -0,0 +1,109 @@
+import Page from '@/components/page'
+import { css } from 'linaria'
+
+const pageStyle = css`
+  text-align: center;
+  display: grid;
+  grid-template-columns: 1fr minmax(60%, 61rem) 1fr;
```
calebeby, line 7: @varCepheid can we use `margin: auto` combined with `width` and `min-width` to constrain its width and center it left-to-right instead of setting up a grid? Alternatively you could set up a grid with just one column and use `justify-content: center` to take up the extra space on the left and right without adding extra empty columns. Then we can avoid setting grid-column on the children.

src/routes/about.tsx
```diff
@@ -0,0 +1,109 @@
...
+          Peregrine.
+        </p>
+        <p>
+          If you have any questions about getting your team onboarded to
+          Peregrine or using the app,{' '}
+          <span class={boldText}>
+            please reach out to the Pigmice team&rsquo;s Peregrine lead, Alex
+            Vennebush
+          </span>
+          , at alexv@pigmice.com.
```
calebeby, line 86: Should this be a link?

src/routes/about.tsx
```diff
@@ -0,0 +1,109 @@
...
+const informationStyle = css`
+  font-size: 22px;
+  grid-column: 2;
+`
+
+const sourceStyle = css`
+  padding: 1rem 2rem 0 2rem;
+  font-family: monospace;
+  grid-column: 2;
+`
```
calebeby, line 25: Out of curiosity, why does the text switch to monospace here? I think it might be suitable to leave the font as-is in the text that links to github.

src/routes/about.tsx
```diff
@@ -0,0 +1,109 @@
...
+            frontend (TypeScript with the React and Babel plugins){' '}
+          </a>
+        </p>
+        <p class={sourceText}>
+          {' '}
+          <a href="https://github.com/Pigmice2733/peregrine-backend">
+            backend (Go connected to SQL){' '}
+          </a>
+        </p>
+        <p class={sourceText}>current version: public release 1.0.0</p>
```
calebeby, line 103: @varCepheid The frontend does not currently have numerical versioning. The backend does, but it is not 1.0.0. Is there a reason to introduce a versioning system? If there is, should we have a discussion about that separately from this PR (maybe in a new issue)?

src/routes/about.tsx
```diff
@@ -0,0 +1,109 @@
...
+          Peregrine or using the app,{' '}
+          <span class={boldText}>
+            please reach out to the Pigmice team&rsquo;s Peregrine lead, Alex
+            Vennebush
+          </span>
+          , at alexv@pigmice.com.
+        </p>
+      </div>
+      <div class={sourceStyle}>
+        <p class={boldSourceText}>source code repositories: </p>
```
calebeby, line 90: Should this be a heading instead of a paragraph with special styling?

calebeby (changes requested):
@varCepheid Thanks for making this PR! I've wanted to have some sort of homepage/intro page for a long time!

src/routes/about.tsx
varCepheid: I thought it looked cooler. Perhaps that's a bad reason.

src/routes/about.tsx
varCepheid: I threw that in because I know there had been discussions about versioning in the past, and I just wanted a placeholder. I think a date-version system would work better, updated whenever we push to the `master` branch.

src/routes/about.tsx
```diff
@@ -0,0 +1,101 @@
...
+            frontend (TypeScript with the React and Babel plugins){' '}
+          </a>
+        </p>
+        <p>
+          {' '}
+          <a href="https://github.com/Pigmice2733/peregrine-backend">
+            backend (Go connected to SQL){' '}
+          </a>
+        </p>
+        <p>current version: are we doing versioning?</p>
```
calebeby, line 95: Can we delete this for now and move the conversation about versioning to another issue?

src/routes/about.tsx
calebeby: Personally, I am not a huge fan, but that's subjective :smile: 

[Here's an article that talks about how it's beneficial to use a consistent font throughout a design](https://elearningindustry.com/role-of-typography-in-web-design-tips-and-best-practices). In my opinion, since both the list of github repos and the text above it would both be considered "body text", I think they should be the same font.

Take that as you will, not trying to force your hand here :wink:

src/routes/about.tsx
```diff
@@ -0,0 +1,101 @@
...
+          , allowing team members to spend time analyzing data and making
+          decisions instead of crunching numbers. It shows an analysis table
+          where team members can compare individual statistics or overall
+          performance between teams and has pages for each team at each event
+          and the matches each team will play in, easily allowing pit crews to
+          know when they&rsquo;re up next. All event data and some match scoring
+          data is pulled from The Blue Alliance.
+        </p>
+        <p>
+          <strong>It&rsquo;s really easy to get started with Peregrine.</strong>{' '}
```
calebeby, line 63: @varCepheid To avoid the annoying spaces-in-embedded-expressions things, we can use an embedded string for the whole text. It would also handle the HTML escaping automatically for you if you do that.

Like this:

```
<strong>{"It's really easy to get started with Peregrine."}</strong>
```
and

```
{" Have a team captain or coach send an email to "}
```

I think that makes it a little easier to read the code and it shouldn't affect the output.

I'd suggest making this change throughout the whole about text.

calebeby (commented):
@varCepheid Thanks for making this about page, it is definitely something that I've wanted to exist for a long time but was never able to prioritize. This PR is getting pretty close to being ready!

src/routes/about.tsx
```diff
@@ -0,0 +1,101 @@
+import Page from '@/components/page'
+import { css } from 'linaria'
+
+const pageStyle = css`
+  text-align: center;
+  padding: 0 1rem;
+  margin: auto;
+  max-width: 50rem;
```
calebeby, line 8: Would you consider reducing the max-width here to something a little narrower? Even though it will make the page taller, it will also make it easier to read.

A good typography rule of thumb that I usually use is to make the lines of text about 80 characters wide at the widest. That makes the text easier to scan and it makes it easier to find the next line when reading.

[Here's more about that](https://kickpoint.ca/the-readability-formula-making-your-website-easy-to-read/)

src/routes/about.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </strong>
+          , at <a href="mailto:alexv@pigmice.com">alexv@pigmice.com</a>.
+        </p>
+      </section>
+      <section class={sourceStyle}>
+        <h3>source code repositories:</h3>
+        <p>
+          {' '}
+          <a href="https://github.com/Pigmice2733/peregrine-frontend">
+            frontend (TypeScript with the React and Babel plugins){' '}
```
calebeby, line 86: Sorry this is a little nitpicky but this is how I would describe it
```suggestion
            frontend (TypeScript and Preact){' '}
```

src/routes/about.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </a>
+        </p>
+        <p>
+          {' '}
+          <a href="https://github.com/Pigmice2733/peregrine-backend">
+            backend (Go connected to SQL){' '}
```
calebeby, line 92:
```suggestion
            backend (Go with Postgres){' '}
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+          {
+            'Go. Peregrine was originally built by students on team 2733 Pigmice in '
+          }
+          {
+            '2017 and redesigned in the autumn of 2018. Development is ongoing to '
+          }
+          {
+            'make the website better than ever. Peregrine has been used by 15 teams '
+          }
+          {'and counting to simplify scouting at FRC competitions. '}
```
calebeby, line 55: Would the code here read easier as a template string? (the adjacent spaces/newlines  will get collapsed into a single space automatically)

Something like:

```ts
{
`
The frontend is written in TypeScript and the backend is written in Go.
Peregrine was originally built by students on team 2733 Pigmice in 2017
and redesigned in the autumn of 2018.
Development is ongoing to etc...
`
}
```

(same for large blobs of text below)

Also, I believe the space at the end of the last sentence here is not needed since it is the end of a paragraph.

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+            }
+          </strong>
+          {'at '} <a href="mailto:alexv@pigmice.com"> {'alexv@pigmice.com.'}</a>
+        </p>
+      </section>
+      <section class={sourceStyle}>
+        <h3> {'source code repositories:'} </h3>
+        <p>
+          <a href="https://github.com/Pigmice2733/peregrine-frontend">
+            {' frontend (TypeScript and Preact) '}
```
calebeby, line 105: I don't think we need the extra spaces at the start and end here since it is the only text in a paragraph. Then we also don't need the string embed:

```suggestion
            frontend (TypeScript and Preact)
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+          </a>
+        </p>
+        <p>
+          <a href="https://github.com/Pigmice2733/peregrine-backend">
+            {' backend (Go with Postgres) '}
```
calebeby, line 110: Same as above:

```suggestion
            backend (Go with Postgres)
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+        </p>
+        <p>
+          {'If you have any questions about getting your team onboarded to '}
+          {'Peregrine or using the app, '}
+          <strong>
+            {
+              "please reach out to the Pigmice team's Peregrine lead, Alex Vennebush,"
+            }
+          </strong>
+          {'at '} <a href="mailto:alexv@pigmice.com"> {'alexv@pigmice.com.'}</a>
```
calebeby, line 98: Formatting nitpick (several changes here):

```suggestion
          {' at '}
          <a href="mailto:alexv@pigmice.com">alexv@pigmice.com</a>.
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+            'performance between teams and has pages for each team at each event '
+          }
+          {
+            'and the matches each team will play in, easily allowing pit crews to '
+          }
+          {"know when they're up next. All event data and some match scoring "}
+          {'data is pulled from The Blue Alliance. '}
+        </p>
+        <p>
+          <strong> {"It's really easy to get started with Peregrine. "}</strong>
```
calebeby, line 76: Formatting nitpick:

```suggestion
          <strong>It's really easy to get started with Peregrine.</strong>
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+          {'Have a team captain or coach send an email to '}
```
calebeby, line 77:
```suggestion
          {' Have a team captain or coach send an email to '}
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <a href="mailto:alexv@pigmice.com"> {'alexv@pigmice.com '} </a>
```
calebeby, line 78:
```suggestion
          <a href="mailto:alexv@pigmice.com">alexv@pigmice.com</a>
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+          {
+            'and a member of the Pigmice will add your team to the dropdown on the '
```
calebeby, line 80:
```suggestion
            ' and a member of the Pigmice will add your team to the dropdown on the '
```

src/routes/about.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <strong>
+            {
+              "please reach out to the Pigmice team's Peregrine lead, Alex Vennebush,"
+            }
+          </strong>
+          {'at '} <a href="mailto:alexv@pigmice.com"> {'alexv@pigmice.com.'}</a>
+        </p>
+      </section>
+      <section class={sourceStyle}>
+        <h3> {'source code repositories:'} </h3>
```
calebeby, line 102:
```suggestion
        <h3> {'Source code repositories:'} </h3>
```

calebeby (changes requested):
Left some minor comments, looking pretty good

src/routes/about.tsx
varCepheid: This doesn't work because of the escape character. I think the way I have it is best.

src/routes/about.tsx
```diff
@@ -0,0 +1,107 @@
...
+            performance between teams and has pages for each team at each event
+            and the matches each team will play in, easily allowing pit crews to
+            know when they're up next. All event data and some match scoring
+            data is pulled from The Blue Alliance.
+          `}
+        </p>
+        <p>
+          <strong>{"It's really easy to get started with Peregrine."}</strong>
+          {' Have a team captain or coach send an email to '}
+          <a href="mailto:alexv@pigmice.com"> alexv@pigmice.com </a>
```
calebeby, line 68: The space is linked here, which I think was not intended:

image

```suggestion
          <a href="mailto:alexv@pigmice.com">alexv@pigmice.com</a>
```

(not sure why the space after is linked but not the space before. Usually for HTML elements you do not want to have spaces right inside of the element declaration, so I'd suggest making this kind of change throughout this file.

src/routes/about.tsx
```diff
@@ -0,0 +1,107 @@
...
+            If you have any questions about getting your team onboarded to
+            Peregrine or using the app, 
+          `}
+          <strong>
+            {
+              "please reach out to the Pigmice team's Peregrine lead, Alex Vennebush, "
+            }
+          </strong>
+          {'at '}
+          <a href="mailto:alexv@pigmice.com"> alexv@pigmice.com. </a>
```
calebeby, line 87: Nit: The ending period is linked, which is not part of the email:

image

```suggestion
          <a href="mailto:alexv@pigmice.com">alexv@pigmice.com</a>{'.'}
```
(not sure how prettier will want to format this)

calebeby (approved):
Couple more tiny suggestions. I am happy with this, feel free to merge once you've addressed these!

calebeby (approved):

PR #1534, opened 2023-11-09
add and edit matches manually

varCepheid:
Admins and super-admins can create new matches for an event manually and edit existing matches. The editable fields are name, date & time, and team list.

### DIFF
--- a/src/components/match-card.tsx
+++ b/src/components/match-card.tsx
@@ -5,6 +5,8 @@ import Card from '@/components/card'
 import { css } from 'linaria'
 import { memo } from '@/utils/memo'
 import clsx from 'clsx'
+import IconButton from './icon-button'
+import { mdiPencil } from '@mdi/js'
 
 interface MatchCardProps {
   match: {
@@ -17,6 +19,7 @@ interface MatchCardProps {
   eventKey: string
   link?: boolean
   class?: string
+  isAdmin?: boolean
 }
 
 const matchCardStyle = css`
@@ -83,7 +86,7 @@ const blueStyle = css`
 `
 
 export const MatchDetailsCard = memo(
-  ({ match, eventKey, link, class: className }: MatchCardProps) => {
+  ({ match, eventKey, link, class: className, isAdmin }: MatchCardProps) => {
     const matchName = formatMatchKey(match.key)
 
     const createTeamLinks = (teams: string[]) =>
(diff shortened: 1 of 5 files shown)


varCepheid:
@calebeby When I try to test the match-creator form in the deploy preview, it throws a red Alert popup, but it doesn't tell me what the problem is. Can you look through the code and tell me if there's anything that should cause a problem?

varCepheid:
The same error occurs when saving the match-editor form.

calebeby:
@varCepheid When I look in devtools it shows me that the PATCH request did not go through; the server returned a `405 Method Not Allowed`:

image

Based on the [API Docs for the backend](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/Pigmice2733/peregrine-backend/develop/internal/server/openapi.yaml), it looks like the server does not support updating matches, only creating new ones. So we are missing a backend method that you (or someone) will need to send a PR for.

image

calebeby:
Looks like there is an undocumented `PUT` for updating matches. Maybe you can use that instead of `PATCH`: https://github.com/Pigmice2733/peregrine-backend/blob/develop/internal/server/routes.go#L40C26-L40C26

And the `POST` method for `/events/{eventKey}/matches` in the documentation is not implemented. You should be able to use the `PUT` directly to make a match instead of using the `POST`.

Maybe it would be good to update the OpenAPI docs in https://github.com/Pigmice2733/peregrine-backend/blob/develop/internal/server/openapi.yaml

varCepheid:
The forms are going through fine now. I've fixed all the problems I could find except that when I go into the match editor, it will always display (in the `placeholder` attribute) the current date-time instead of the match's date-time. It should only do that if it is unable to load the match or the match's `time` property in the code, but it's weird to me that it wouldn't be able to load that information, since it displays correctly on the match card. Any ideas?

calebeby:
@varCepheid It is because the `match` variable is `null` for a bit while `useMatchInfo` is fetching the match info. During that time, `matchDate` gets set to the current time, and then `day` and `time` get initialized with those values. You did not write any code to update `day` and `time` when `matchDate` changes (/ is loaded). You could use a [`useEffect`](https://preactjs.com/guide/v10/hooks/#useeffect) for this?

varCepheid:
Added the `useEffect` call and fixed the problem that I mentioned in issue #1536.

PR #1539, opened 2024-03-05
push to production-version 2024

varCepheid:
Changes since the last push are listed in the commits.


varCepheid:
Almost as good as the Vite PR!

calebeby:
If we merge this PR as "rebase and merge", then it will not add a merge commit to the `main` branch. That way main will not be any commits ahead of `dev`.

calebeby (approved):

PR #1545, opened 2024-05-07
add arrow on scouting form dropdown menu

### LINKED ISSUE #1541: dropdown arrow for match selection on scouting page doesn't appear on certain devices
There is supposed to be an arrow next to the name of the match (Qual 2 in this image). It appears fine on other devices. Possibly, but not definitely, an Apple/Android issue.

varCepheid:
closes #1541 
Forces all dropdown elements to have an arrow that tells the user that the element is a dropdown.

### DIFF
--- a/src/components/dropdown.tsx
+++ b/src/components/dropdown.tsx
@@ -9,6 +9,7 @@ const dropdownStyle = css`
   font-size: 0.8rem;
   color: var(--off-black);
   font-weight: bold;
+  appearance: menulist;
 
   & option {
     text-transform: none;


calebeby:
If you could show me a before/after screenshot on iOS that would be great! I can't reproduce the bug on Android. Then this should be good to go!

calebeby:
@varCepheid did your fixes work?

varCepheid:
Nope, nothing has worked, but I also don't have an Apple device, so it's difficult for me to test changes. I'll mark this so it's clear.

PR #1547, opened 2024-08-22
Add a checkmark for played matches

### LINKED ISSUE #1544: Add a check mark on match cards if the match has been played (if it has a score)
I don't think there is space to put the actual score, but at least a check mark would be helpful. I think modifying the match card itself so that this change appears everywhere that it is used would be great.

I am imagining a small check icon on the left of the word "Qual" here:

image

I like this one, maybe in grey?

image

Or maybe on the right side of the time (or instead of the time, but then you can't see when matches were played?)

Here are some places where it would be especially helpful:

General match list
image

Match list for specific team

image

varCepheid:
On the match cards for played matches on the event and event-team-matches pages, there is a checkmark between the name of the match and the timestamp. Ideally, the name appears close to the checkmark in all circumstances.
closes #1544

### DIFF
--- a/src/components/match-card.tsx
+++ b/src/components/match-card.tsx
@@ -20,7 +20,6 @@ interface MatchCardProps {
   eventKey: string
   link?: boolean
   class?: string
-  checkmark?: boolean
 }
 
 const matchCardStyle = css`
@@ -34,11 +33,12 @@ const matchCardStyle = css`
   & > time {
     grid-row: span 2;
     grid-column: 3;
-    place-self: center end;
+    place-self: center center;
     font-size: 0.85rem;
     color: var(--grey-text);
     white-space: nowrap;
     text-overflow: ellipsis;
+    margin-left: 0.3rem;
   }
 `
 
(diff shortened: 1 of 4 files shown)


varCepheid:
I can't really test that the changes work aesthetically when the match hasn't been played, since I couldn't find any events with matches that hadn't been played yet, but it looks good to me on both pages where the checkmark is present and where it isn't.

src/components/match-card.tsx
```diff
@@ -5,30 +5,35 @@ import Card from '@/components/card'
...
     redAlliance: string[]
     blueAlliance: string[]
     time?: Date
+    redScore?: number | undefined
   }
   key?: string | number
   eventKey: string
   link?: boolean
   class?: string
+  checkmark?: boolean
```
calebeby, line 23: @varCepheid It looks like the only place where this is _not_ set is on event-match. Is there a reason to not show it there? It seems like it might simplify it to always show the checkmark on played match cards, regardless of where in the UI you are. What do you think?

calebeby (dismissed):
Something about me is slightly bothered that they don't quite line up vertically :joy:. Not a big deal, feel free to fix if it's easy (maybe put it on the other side?), or leave it as is if it's not easy.
image
Also, here is an alternative color (same as the time text): 
image

Not sure which I like better. The grey makes it a little more subtle compared to the match name, which I like. Either way is fine with me.
Feel free to merge without another review. Thanks for spending time on this PR!

varCepheid:
I tried really hard to make the grid look the same across all the cards, and I just couldn't get it to work. I did the gray checkmarks.

PR #1549, opened 2024-09-29
sort realms alphabetically in dropdown on signup page

### LINKED ISSUE #1548: sort realms alphabetically in sign-up dropdown

varCepheid:
closes #1548 
Title says it all.

### DIFF
--- a/src/routes/signup.tsx
+++ b/src/routes/signup.tsx
@@ -49,6 +49,8 @@ const SignUpForm = () => {
   const realms = usePromise(getRealms) || []
   const emitError = useErrorEmitter()
 
+  realms.sort((a, b) => a.name.localeCompare(b.name))
+
   const onSubmit = (e: Event) => {
     e.preventDefault()
     if (realmId === undefined) return
--- a/src/routes/signup.tsx
+++ b/src/routes/signup.tsx
@@ -49,8 +49,6 @@ const SignUpForm = () => {
   const realms = usePromise(getRealms) || []
   const emitError = useErrorEmitter()
 
-  realms.sort((a, b) => a.name.localeCompare(b.name))
-
   const onSubmit = (e: Event) => {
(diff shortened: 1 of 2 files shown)


varCepheid:
There are some changes to `package-lock.json` because I installed a newer patch of NPM. Would you like me not to do that in the future?

src/routes/signup.tsx
```diff
@@ -49,6 +49,8 @@ const SignUpForm = () => {
   const realms = usePromise(getRealms) || []
   const emitError = useErrorEmitter()
 
+  realms.sort((a, b) => a.name.localeCompare(b.name))
```
calebeby, line 52: Let's avoid running this sort on every render. It is against the rules of hooks to modify variables directly in the component body: https://react.dev/reference/rules/components-and-hooks-must-be-pure.

Breaking the rules of hooks doesn't mean that your code won't work, but it does mean that future changes to react/preact might break your code even if it works now.

Since `sort()` mutates the underlying variable, probably better to do chain it on to the promies so it just happens once:

```
usePromise(() => getRealms().then(realms => realms.sort(...)) || []
```

You could also modify getRealms function directly so that it always returns a sorted list. Probably would be a worthwhile change.

varCepheid:
I noticed that getRealms is also being called in the report-editor page and the users page. In the users table, the user can sort by any column of the table, so the order it gets the realms is unimportant. On the report-editor page, the only place where it uses the realms is if the user is a super-admin, in which case it gives an option to replace the user who created the report with any other user in any realm, using a dropdown that groups users by realm. It doesn't seem to be sorting the groups alphabetically, but the code is a little over my head, and it seems like a very niche use-case. Do you want me to try to address that?

calebeby:
I think this PR is good as-is, thanks for making the suggested change!

calebeby (approved):

